PR #60, opened 2014-10-28
Added test coverage support using instanbul.js

Mickael-van-der-Beek:
This pull request adds test coverage support with `istanbul` and `Grunt.js`. Imported modules are:

- `istanbul`, for `LCOV` format report generation
- `grunt`, task runner
-  `grunt-cli`, run locally installed `Grunt.js` version
- `load-grunt-tasks`, practical util that imports task providers from NPM modules
- `grunt-istanbul`, task that runs the `istanbul` NPM module on the spec,
-  `grunt-jasmine-node`, task that runs the `jasmine-node` NPM module to launch the tests
- `grunt-contrib-copy`, task that copies the `package.json` file needed by the `RestClient.js` file into the `coverage` directory so as not to break the relative `require()` path
- `grunt-contrib-clean`, task that deletes the previous test coverage `reports` and `instruments`

Even though the Twilio team itself doesn't use test coverage, I thought that it could be a cool addition.
In my opinion, test coverages make it easier for new-comers to contribute to the project.

If you guys like the idea, I can also create a second pull-request that would add support for a badge (in the `README.md` file) and test coverage visualisation using [Coveralls](https://coveralls.io/). Coveralls is free because the project is open-source. You only need to GitHub-connect to the site.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 .DS_Store
 config.js
 node_modules
+spec/coverage
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+before_script:
+  - npm update -g npm
 node_js:
     - "0.10"
     - "0.8"
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,0 +1,59 @@
+function GruntTasks (grunt) {
+	'use strict';
+
+	grunt.initConfig({
+		clean: {
+			coverage: [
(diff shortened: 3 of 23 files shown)


kwhinnery:
thanks for this, but the hope is to do away with the jasmine-node test suite and use something like tape or mocha which has some community support around it. We could look at doing a coverage tool after that happens. I did go ahead and pull in your other PR and have recently updated the travis config to sidestep the issue you mentioned here.

PR #71, opened 2015-01-09
empty body isn't an error condition :S - fixes #63

### LINKED ISSUE #63: Media "delete" returns no error, but data has ``status: 500``.
I'm running this:

    var twilioClient = require("twilio")(<AccountSID>, <AuthToken>);
    twilioClient.messages(<MessageID>).media(<MediaID>).delete(function(err, data) {
          console.log(err, data); // Logs "null, {status: 500, message: 'Empty body'}"
    });

Within the callback, ``err`` is null, but data is ``{status: 500, message: "Empty body"}``.  The media file is successfully deleted.  It's confusing to get a reported 500 status when the command succeeds and no ``err`` is given.

kwhinnery:
Remove incorrect error data from the data object. Returning an empty body is not an error condition.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,3 +2,5 @@ language: node_js
 node_js:
     - "0.10"
     - "0.8"
+before_install:
+- sudo npm install npm -g
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ node_js:
     - "0.10"
     - "0.8"
 before_install:
-- sudo npm install npm -g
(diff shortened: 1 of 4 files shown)


skimbrel:
regression test?

kwhinnery:
Still needs one - fixing current tests.

kwhinnery:
@skimbrel should be gtg - if you want to sanity check it right quick that would be great.

skimbrel:
+1

PR #72, opened 2015-01-09
Remove Media resource

kwhinnery:
Account-level media doesn't exist in the API yet, remove.


kwhinnery:
Travis will pass with changes from https://github.com/twilio/twilio-node/pull/71

PR #73, opened 2015-01-21
add dateSent to list of date converted strings

christiangenco:
Right now the 'dateSent' property isn't getting automatically formatted as a date. This field is returned in the `getSms` response callback.

I'd like to add that this method of converting dates should be reconsidered for something more flexible. I'd propose searching *all* fields for things that look datey, or convert to a valid date when they're `Date.parse()`'d. This would add a little processing overhead, though.

PS: I fudged up this pull request by creating another pull request on my own branch, but am unsure how to undo what I did D:

### DIFF
--- a/lib/RestClient.js
+++ b/lib/RestClient.js
@@ -88,7 +88,7 @@ function processKeys(source) {
         });
 
         //Look for and convert date strings for specific keys
-        ['startDate', 'endDate', 'dateCreated', 'dateUpdated', 'startTime', 'endTime'].forEach(function(dateKey) {
+        ['startDate', 'endDate', 'dateCreated', 'dateUpdated', 'startTime', 'endTime', 'dateSent'].forEach(function(dateKey) {
             if (source[dateKey]) {
                 source[dateKey] = new Date(source[dateKey]);
             }


kwhinnery:
Thanks for this! Will go ahead and merge.

I would be down for something more intelligent, but I'm not sure what that looks like. I don't want to try and cast every string in a JSON payload, both for processing overhead and the possibility of mistakenly creating a valid date object with string content that's not really a date. Nor do I like the idea of keywords in JSON keys or the like, as the API could add a field that has certain words in it but is not Date convertible.

I am open to suggestions on that front though and would consider a PR with such a feature added.

christiangenco:
Neat! Thanks :D

Once it's pushed through to npm and atmosphere I can delete four lines of code in my project :wink: 

Agreed on overhead for casting every string. What if we just tested strings against some regex? 

The Twilio date format looks like "Wed, 21 Jan 2015 22:51:45 +0000", so maybe `/^\w+,\s\d+\s\w+\s\d{4}\s\d{2}:\d{2}:\d{2}\s\+\d{4}$/`?

philnash:
Great idea, there's a middleware parser in Ruby that does this for dates, the main part of which can be seen [here](https://github.com/lostisland/faraday_middleware/blob/master/lib/faraday_middleware/response/parse_dates.rb). It too uses a regex to check things that look like dates and parse them.

christiangenco:
Oh neat! That can't be an awful way of doing it, then.

I'll throw something together and shoot over another pull request.

christiangenco:
Just submitted a new pull request! https://github.com/twilio/twilio-node/pull/74

PR #74, opened 2015-01-23
Add dynamic date converstion for any date-like value

christiangenco:
This will convert any string value that matches the regex `/^\w+,\s\d+\s\w+\s\d{4}\s\d{2}:\d{2}:\d{2}\s\+\d{4}$/` to a date with `new Date(value)`.

Note that this diverges from the current functionality by converting both the original `underscored_key` and the aliased `camelCasedKey` (the previous version only converted the later).

Also: I couldn't for the life of me create a failing test in the current test framework. I noted that there was an open issue to switch to mocha, so I wrote my tests in that instead. They can be run with `mocha` in the project directory.

### DIFF
--- a/lib/RestClient.js
+++ b/lib/RestClient.js
@@ -66,11 +66,16 @@ function RestClient(sid, tkn, options) {
 }
 
 //process data and make available in a more JavaScripty format
-function processKeys(source) {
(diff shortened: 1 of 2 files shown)


christiangenco:
Ooo, mocha should be integrated in the Travis CI builds. I'm not sure how that works: does Travis just run `npm test`?

christiangenco:
bump

jingming:

PR #86, opened 2015-04-19
Fix an escaping bug we had

A-Zak:
querystring was changing pluses '+' to spaces ' ' and changing it to qs fixed it for us... :)

### DIFF
--- a/lib/Client.js
+++ b/lib/Client.js
@@ -1,6 +1,6 @@
 //Dependencies
 var Q = require('q');
-var querystring = require('querystring');
+var qs = require('qs');
 var request = require('request');
 var moduleinfo = require('../package.json');
 var _ = require('underscore');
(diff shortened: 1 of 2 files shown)


skimbrel:
+1. @kwhinnery, any gotchas to look out for here?

A-Zak:
Don't think so... The weird plus sign escaping was the only thing we saw...

kwhinnery:
@skimbrel @A-Zak the reason we are using `querystring` here is because we prefer serializing arrays differently. To work with the Twilio REST API, we need for:

```javascript
var params = { MediaUrl: ['foo', 'bar', 'baz'] };
```

to serialize to

```
MediaUrl=foo&MediaUrl=bar&MediaUrl=baz
```

`qs` automatically adds array notation subscript to parameters like this, so the above becomes

```
MediaUrl[0]=foo&MediaUrl[1]=bar&MediaUrl[2]=baz
```

@A-Zak can we back up to the original problem (replacing spaces with "+") and work a solution from there?

kwhinnery:
Also, we probably need a test for this, if the CI passed on this PR that's bad :(

A-Zak:
Ok, got it... we'll try to find some way around it...
Glad I could've at least help finding the missing test case... :)
Cheers...

Issue #97, opened 2015-05-28
Pass Cookies in sendMessage method


trever:
I'm looking to pass a `headers['Set-Cookie']` value into the method's option argument.

Use Case: App sends an hourly text message to a user and expects a response. On the cron job's execution (sending the SMS), it should tag that message with a "session identifier". On the next job execution it creates a new session. That way, if the user fails to respond to the first SMS, but does respond to the second, session integrity is maintained. 

Thoughts?

skimbrel:
Unfortunately this functionality isn't available in the Twilio REST API. Cookies are supported by the Twilio TwiML engine, which I'm guessing is what you saw in our documentation (https://www.twilio.com/docs/api/twiml/sms/your_response), but you'll need to have an inbound message from the user first to set a cookie on the session.

trever:
Got it. Yeah, I was referring to this - https://www.twilio.com/blog/2014/07/the-definitive-guide-to-sms-conversation-tracking.html but didn't read toward the bottom where it says exactly what you mentioned.

I'll close now.

Thanks!

PR #98, opened 2015-05-28
Support headers object in Twilio HTTP Calls

trever:
Referencing issue #97


jmctwilio:
This PR is very out-of-date.  It will likely be subsumed by a new update that exposes the last request and response objects (if it hasn't already been addressed).  If those new changes are insufficient and the functionality is necessary, please let us know.  I'm very sorry to have left this PR open for so long, but there was an initiative to move the code over to a generative model which took precedence.  The team should be less overwhelmed as a result and more responsive to future PRs and Issues.

Issue #96, opened 2015-05-28
If you pass mediaUrl:undefined to the sendMessage function, no call is made


jonathankeebler:

skimbrel:
Can you clarify a bit more? By "no call is made" do you mean the client doesn't emit an HTTP request to Twilio, or that Twilio doesn't do anything in response?

jonathankeebler:
Traced it a bit more and I'm getting this reply from Twilio API: 
```
{ status: 400,
  message: 'Media urls:  are invalid. Please use only valid http and https urls',
  code: 21620,
  moreInfo: 'https://www.twilio.com/docs/errors/21620' }
```

I'm working on a patch.

skimbrel:
What would you like to patch in the library for this? The library shouldn't be responsible for validating all parameters of all REST requests, as that's already implemented in the actual API handler.

jonathankeebler:
It was a feature not a bug ;)

PR #100, opened 2015-05-28
Ensure a param is set before sending it in the request

jonathankeebler:
What about this simple change to resolve? #96


skimbrel:
There are some endpoints that allow setting parameters to empty values to delete fields from a resource. For example, you can redact the body of a stored Message record by POSTing to its instance URL with "Body=" in the form, or you can delete an SmsUrl/VoiceUrl from an Application by POSTing an empty string for that parameter. 

I appreciate the effort, but it probably makes more sense for you to check for this case in your own application code.

jonathankeebler:
Fair enough.

Would you consider adding some Mocha test cases? I created some to test this out.

PR #104, opened 2015-07-01
Updates Twilio dashboard screengrab to gif.

philnash:
This is a more up to date gif, although things have changed a bit more since this was taken. Still, I think this is probably better than the existing screengrab.


skimbrel:
+1

PR #105, opened 2015-07-15
fix validateExpressRequest to allow query params in webhook url

byronmwong:


ethanresnick:
I think the implementation in this PR is buggy, though, in that `req.path` doesn't always contain the full pathname (http://stackoverflow.com/a/39827135). See my alternative implementation in #194 (which I opened before seeing this; sorry!)

jmctwilio:
This PR is very out-of-date.  If it hasn't already been addressed, if our new changes are insufficient and the functionality is necessary, please let us know. I'm very sorry to have left this PR open for so long, but there was an initiative to move the code over to a generative model which took precedence. The team should be less overwhelmed as a result and more responsive to future PRs and Issues.

PR #116, opened 2015-08-12
Add sub node TaskAttributes to Enqueue node

metherton:
Whilst doing the TaskRouter getting started tutorial (in node) I found that it was not possible to do the step requiring an Enqueue node in the response as the TaskAttributes was not recognised as a valid sub node of the Enqueue node. I therefore added it as a legal sub node in the TwimlRepsponse class.


jingming:

Issue #117, opened 2015-08-28
TypeError: undefined is not a function


danielckv:
After npm install a few days ago, our app started to crash each time. I tried to figure out what is the problem but didn't find any solution for this:

at Function.generate.restFunctions (/node_modules/twilio/lib/resources/generate.js:89:50)
    at module.exports (/node_modules/twilio/lib/resources/ListInstanceResource.js:29:14)
    at mixinResources (/node_modules/twilio/lib/resources/Accounts.js:18:31)
    at module.exports (/node_modules/twilio/lib/resources/Accounts.js:109:5)
    at new RestClient (/node_modules/twilio/lib/RestClient.js:28:58)
    at initializer (/node_modules/twilio/lib/index.js:18:12)

When I removed the 'toLowerCase' method from the generate.js file it worked.
Any solution for this problem?

westy92:
The referenced PR has been merged.  Can this issue be closed?

PR #118, opened 2015-08-30
Fixed issue in generate.js when trying to attach undefined object.

danielckv:
After a few days of searching, i found finally the issue of this issue: https://github.com/twilio/twilio-node/issues/117

[Review on Reviewable](https://reviewable.io/reviews/twilio/twilio-node/118)

### DIFF
--- a/lib/resources/generate.js
+++ b/lib/resources/generate.js
@@ -86,7 +86,9 @@ generate.restFunctions = function (object, client, methods, resource) {
         else {
             //Create an alias for the given method name to a REST function
             for (var key in method) {
-                object[key] = object[method[key].toLowerCase()];
+            	// fixed issue for null Properties.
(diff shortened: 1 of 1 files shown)


okev:
Any progress on this? Looks like its still in master and the latest NPM version (2.9.0).

Tuetuopay:
Can someone please put this into the NPM repository ? I have the same issue (though I fixed it locally), but it would be better to have the fix through NPM

okev:
If it helps @Tuetuopay my approach was to fork, fix and npm install from that repo. I've filed a support ticket with Twilio directly, but no response yet.

Tuetuopay:
@okev that's exactly what I ended up doing !

carlosdp:
Thanks!

PR #128, opened 2015-10-16
return the result of get in getSms, getMessage, and getCall

srveit:


jingming:

PR #131, opened 2015-10-27
Fix flaky client test

Ragil:
1. Use local server to validate request timeouts
2. Upgrade jasmine-node to 2.0.x

### DIFF
--- a/spec/client.core.spec.js
+++ b/spec/client.core.spec.js
@@ -1,110 +1,155 @@
 var twilio = require('../index');
+var request = require('request');
+var moduleinfo = require('../package.json');
+var http = require('http');
 
-describe('The Twilio REST Client constructor', function () {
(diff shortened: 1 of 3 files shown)


spec/client.core.spec.js
```diff
@@ -1,110 +1,155 @@
...
+            var specialClient = new twilio.RestClient(fakeSid, fakeAuth, {
+                host: fakeHost,
+                apiVersion: fakeApi
+            });
+
+            //Check URL construction with custom config
+            expect(specialClient.getBaseUrl()).toBe('https://' + fakeSid + ':'
+                    + fakeAuth + '@' + fakeHost + '/' + fakeApi); });
+
+        xit('should fail gracefully if the host is unreachable', function (done) {
```
jingming: xit?

PR #132, opened 2015-10-27
Fix flaky client tests

Ragil:
1. Use local http server to verify client timeout.
2. Bump jasmine-node to version 2.0.x.


jingming:
+1

PR #134, opened 2015-11-16
Updates docs to use bodyParser instead of express.urlencoded

philnash:
Adds a couple of other things too:

* 403 response when failing to validate the request
* splits up the two examples of `validateExpressRequest` so they are more obvious
* adds `app.listen` line to start server
* adds spaces between `//` and comments


carlosdp:
:+1:

PR #136, opened 2015-12-09
Expose update and delete message wrappers

senthgit:
Message - Update and Delete Message


lib/resources/ip_messaging/Services.js
```diff
@@ -37,7 +37,7 @@ module.exports = function(client) {
             channelApi.messages = NextGenListResource(
                 client,
                 baseUrl + '/' + sid + '/Channels/' + channelSid + '/Messages',
-                ['GET'],
+                ['GET', 'DELETE', {update: 'POST'}],
```
Ragil: you need to specify 'POST' as well, otherwise it will only have update

senthgit: added POST as well.

Ragil:
+1

PR #142, opened 2016-01-12
RestClient getSms, getMessage and getCall now return promises

sheppardv:
So actual `RestClient` `getSms`, `getMessage` and `getCall` now return promises.
Tests added.


kwhinnery:
+1

jingming:

PR #147, opened 2016-01-21
Loading Reservations under a Worker by default in TaskRouter JS SDK

jwitz10:
- Allow fetching Reservations under a Worker by default
- Allow updating Reservations under a Worker

### DIFF
--- a/lib/TaskRouterWorkerCapability.js
+++ b/lib/TaskRouterWorkerCapability.js
@@ -4,12 +4,14 @@ var TaskRouterCapability = require('./TaskRouterCapability');
 function TaskRouterWorkerCapability(accountSid, authToken, workspaceSid, workerSid) {
     TaskRouterCapability.call(this, accountSid, authToken, workspaceSid, workerSid);
 
-    this.reservationsUrl = this._baseUrl + "/Tasks/**";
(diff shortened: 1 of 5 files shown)


jingming:
should we add a test for `allowReservationUpdates`?

jingming:
+1

lib/TaskRouterWorkerCapability.js
```diff
@@ -4,12 +4,14 @@ var TaskRouterCapability = require('./TaskRouterCapability');
 function TaskRouterWorkerCapability(accountSid, authToken, workspaceSid, workerSid) {
     TaskRouterCapability.call(this, accountSid, authToken, workspaceSid, workerSid);
 
-    this.reservationsUrl = this._baseUrl + "/Tasks/**";
+    this.tasksUrl = this._baseUrl + "/Tasks/**";
```
jingming: same here, if I had an instance of `TaskRouterWorkerCapbility` that accessed `tasksUrl`, it would break with this update.

jwitz10: Correct, but I'm not sure anyone would directly reference reservations_url since this variable's only purpose would be for generating access policies.

I can revert this if necessary though.

jingming:
+1

PR #156, opened 2016-03-15
add option for static jwt

GregPrice24:


jmctwilio:
This PR is very out-of-date.  If our recent changes are insufficient and the functionality is necessary, please let us know. I'm very sorry to have left this PR open for so long, but there was an initiative to move the code over to a generative model which took precedence. The team should be less overwhelmed as a result and more responsive to future PRs and Issues.

PR #160, opened 2016-04-01
Use a supplied fallbackurl if it is a fallback request

pcflmb:
I use a Gateway to pass requests to staging/production servers automatically, unfortunately the Gateway has a propensity to go down every couple months for a few hours, which is unacceptable, so I set my fallback URL to point directly to my production server. To do this, I use the extra logic below so the same function can handle requests via unique request and fallback URLs.

In addition, this includes logic that allows testing the value of the request URL, which is stored as `request.body.ErrorUrl` . Not sure if this is desired or not, but I figured it made sense, and you can just leave `opts.url` blank should you not want that check.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
     - "0.10"
     - "0.8"
 before_install:
+- npm install npm@2 -g
 - npm install npm -g
 notifications:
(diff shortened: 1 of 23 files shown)


jingming:

PR #163, opened 2016-04-26
Throw Error objects instead of strings

kevinburkeshyp:
Throwing values that don't inherit from Error can have unintended consequences.
Among other things, people catching an error may expect a `message` property to
be present on the thrown object. Furthermore, Error objects have a stack trace
attached, which makes it easy to see where the error is thrown from, and what
called the error-generating function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
     - "0.10"
     - "0.8"
 before_install:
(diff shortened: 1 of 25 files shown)


jingming:
The `3.x.x` line of this library throws exceptions that inherit from `Error`.

PR #189, opened 2016-08-18
fix parameter format for hook middleware function

PRE90:


jingming:

PR #194, opened 2016-09-13
Fix validateExpressRequest url auto detection

ethanresnick:
When an incoming express request is being validated (i.e., its signature is being checked), and the URL to use during the check is being determined by inspecting the express request object, the current code fails if the requested url contains query parameters. 

To generate a signature against which to check twilio's provided signature, the code has to figure out what the incoming URL was. And, while doing that, the library currently calls `url.format` and uses `req.originalUrl` as the pathname. (See [here](https://github.com/twilio/twilio-node/pull/194/files#diff-e37ef46c4254746b09275706c1f1976aL45).) However, pathname's aren't allowed to contain query strings so, when `req.originalUrl` contains a query string, `url.format` is being inappropriately provided a query string as part of the pathname. Accordingly, `url.format` escapes the query-string related characters (namely the `?`) to produce a valid pathname. As a result, the url that the library thinks is the incoming URL doesn't match the URL that twilio actually signed, and the signature verification improperly fails. 

To fix this, we parse the `request.originalUrl` (which is a relative URL) to get whatever path, query, etc. stuff it contains, and then merge that in with the protocol and host (which the user may have manually overridden).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
 node_js:
     - "0.10"
     - "0.8"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
     - "0.10"
     - "0.8"
 before_install:
+- npm install npm@2 -g
 - npm install npm -g
 notifications:
   slack:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,10 @@ node_js:
     - "0.8"
 before_install:
 - npm install npm -g
+notifications:
+  slack:
+    on_success: change
+    on_failure: change
+    rooms:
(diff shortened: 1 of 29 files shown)


ethanresnick:
Hi, I've updated the original post to be clearer about the issue and the solution, and also added a test case that fails in the absence of this change. Would it be possible to get this code merged in?

ethanresnick:
I also just rebased the code to make the changes more minimal and explicit, as I can understand hesitation around merging anything that touches the signing logic, and I want to make this change as easy to audit as possible.

ethanresnick:
C'mon twilio... some version of this issue has been open for almost two years, going back to the original PR in #105. Please fix this already!

jmctwilio:
@ethanresnick Hey.  Sorry for the very long delay.  As I'm sure you've seen, the library has undergone a lot of changes.  This and our other helper libraries represented more work than we could keep up with easily.  As a consequence, the team took on an ambitious project to convert the libraries to auto-generated solutions.  With our last library finally being completed, we're able to re-engage with the community.  Please let us know if this continues to be an issue with the new library and feel free to open a PR against it if you would.  Thank you for your contribution and patience.  I believe we can be more responsive now that we've been able to automate much of the code.

PR #195, opened 2016-09-15
Use _.contains instead of _.includes

liamdon:
The package.json for `twilio-node` specifies an `underscore` version of `1.x`. However, the `_.includes` function alias only appeared in `underscore@1.8.x`, so the library is broken for older underscore versions, even those these should be valid according to the package.json.

`_.contains` is a valid alias in all `1.x` versions and so is a better choice to offer the widest compatibility - this is important because underscore may often be included as a peer dependency.


jingming:
👍

jingming:
Thanks! This is released in `2.9.3`.

PR #202, opened 2016-10-18
Minor reformat for readability

piperchester:


jingming:

PR #208, opened 2016-11-08
(Current) Removes 0.8 from CI tests, adds v4, v5, v6 and latest.

philnash:


jingming:
This has been updated in the latest version.

PR #210, opened 2016-12-09
(NextGen) Reusing SSL (TLS) connection and session with forever-agent

magorski:
Twilio REST API is hosted on AWS in Virginia. Response time for me (from UK) is everything in the range of 300-700. It'll be nice if they could have global infrastructure for API so request are routed to servers closest to the originator location (i.e. with Amazon Route 53 [Multi-Region Latency Based Routing](https://aws.amazon.com/blogs/aws/latency-based-multi-region-routing-now-available-for-aws/)).

So currently latency could be a serious issue especially if your workflow requires several calls to the API. But I think I found a optimisation. There is a way to keep connection and SSL session alive between calls so yes, the first connection will still be slow but at least the consecutive ones will be quick. I've used [forever-agent](https://github.com/request/forever-agent) (HTTP Agent that keeps socket connections alive between keep-alive requests). It's implemented within [request](https://github.com/request/request) a http client for [twilio-node](https://github.com/twilio/twilio-node) NextGen therefore this PR is just a one line of code.

So. I wrote a piece of code to test it:
```
var client = require('[Account SID]','[Auth Token]');
var start = null;
var num = 0;
var sum = 0;
var first = 0;
var repeat = 100;
function run() {
    if (num++ < repeat) {
        start = new Date().getTime();
        client.conferences.get('[Existing Conference SID]').fetch(function (err, result) {
            time = new Date().getTime() - start;
            num > 1 ? sum += time : first = time;
            console.log('Run: ' + num + ' | Response time:', ((new Date().getTime() - start) / 1000).toFixed(3), num > 1 ? " (Avg: " + ( sum / (num - 1) / 1000 ).toFixed(3) + ")" : '');
            run();
        });
    } else {
        console.log('Total time: ' + ((first + sum) / 1000).toFixed(3) + ' | Total avg. (including first call): ' + ((first + sum) / repeat / 1000).toFixed(3))
    }
}
run();
```

Before my change the output is:
```
Run: 1 | Response time: 0.607
Run: 2 | Response time: 0.340  (Avg: 0.340)
Run: 3 | Response time: 0.339  (Avg: 0.340)
...
Run: 99 | Response time: 0.339  (Avg: 0.357)
Run: 100 | Response time: 0.328  (Avg: 0.357)
Total time: 35.942 | Total avg. (including first call): 0.359
```

And then after my change it'll be:
```
Run: 1 | Response time: 0.615
Run: 2 | Response time: 0.157  (Avg: 0.156)
Run: 3 | Response time: 0.141  (Avg: 0.148)
...
Run: 99 | Response time: 0.289  (Avg: 0.156)
Run: 100 | Response time: 0.151  (Avg: 0.156)
Total time: 16.063 | Total avg. (including first call): 0.161
```

So you see that the difference is dramatic.

Obviously i'm not an SSL security expert and this one line of code potentially could be somewhere else but I couldn't find a way to pass any http client configuration from twilio initialisation method.

Please comment.

BTW. I've created similar PR (#211) for current version of this library.

BTW. If you ran test code without optimisation on AWS server in Virginia you're getting:
```
Run: 1 | Response time: 0.092
Run: 2 | Response time: 0.058  (Avg: 0.058)
Run: 3 | Response time: 0.076  (Avg: 0.067)
...
Run: 99 | Response time: 0.069  (Avg: 0.070)
Run: 100 | Response time: 0.070  (Avg: 0.070)
Total time: 7.002 | Total avg. (including first call): 0.070
```


jingming:
This change looks fine, but we will have to make it in an internal system. We'll make sure to keep this ticket updated.

PR #216, opened 2017-01-04
Fix invalid signature digest format string

ethanresnick:
Node seems to compare the digest string case-insensitively, so calling
digest with ‘Base64’ and ‘base64’ do the same thing. However, this is
undocumented behavior, and ‘base64’ is the documented way to do this.
See: https://nodejs.org/api/crypto.html#crypto_hmac_digest_encoding

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
 node_js:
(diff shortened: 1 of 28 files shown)


jingming:

PR #219, opened 2017-01-23
Fix accessToken payload by renaming headers to header

ryan-rowland:
@jingming @liamdon @ihumanable 

Now that we're using `jwt.sign` I get fatal errors around `headers` in the payload, as it should be `header`:

```
{ ValidationError: "headers" is not allowed
    at Object.exports.process (/Users/rrowland/git/video-app-service/node_modules/joi/lib/errors.js:140:17)
    at internals.Object.internals.Any.validate (/Users/rrowland/git/video-app-service/node_modules/joi/lib/any.js:667:25)
    at Object.module.exports [as sign] (/Users/rrowland/git/video-app-service/node_modules/jsonwebtoken/sign.js:93:47)
    at AccessToken.toJwt (/Users/rrowland/git/video-app-service/node_modules/twilio/lib/AccessToken.js:188:16)
...
  name: 'ValidationError',
  details:
   [ { message: '"headers" is not allowed',
       path: 'headers',
       type: 'object.allowUnknown',
       context: [Object] } ],
  _object:
   { headers: { cty: 'twilio-fpa;v=1', typ: 'JWT' },
     algorithm: 'HS256',
     issuer: 'SKe44db24afa9185b46afc015586941d9a',
     subject: 'AC96ccc904753b3364f24211e8d9746a93',
     expiresIn: 1800 },
  annotate: [Function] }
```

Will this fix break other things?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: node_js
+sudo: false
 node_js:
     - "0.10"
     - "0.8"
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ node_js:
     - "0.10"
     - "0.8"
 before_install:
+- npm install npm@2 -g
 - npm install npm -g
 notifications:
   slack:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,10 @@ node_js:
     - "0.8"
 before_install:
 - npm install npm -g
+notifications:
+  slack:
+    on_success: change
+    on_failure: change
+    rooms:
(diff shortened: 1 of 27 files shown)


ryan-rowland:
Update: Discovered this was a dependency issue. We are using 5.4.x, which seems incompatible with 7.2.x. In our project, we're using 7.2.1 which was causing a conflict. Seems safe to update to 7.2.1 with this small change; tests are passing.

ryan-rowland:
@kwhinnery Do you think this dependency change is too risky? We have a ticket in Video App to switch to main line twilio-node release an this is currently blocking. Let me know if there are any next steps I should take.

jingming:

PR #225, opened 2017-03-02
Update package.json

kievechua:
When using [cost-of-modules](https://github.com/siddharthkp/cost-of-modules) to inspect my projects.
Found out Twilio is top of it:
│ name        │ children     │ size   │
├─────────────┼──────────────┼────────┤
│ twilio      │ 89           │ 42.45M │
Further investigate show docs directories taken 33.2mb
Can we exclude the doc directories to fasten up install process? Especially for production.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,10 @@
     "type": "git",
     "url": "https://github.com/twilio/twilio-node.git"
   },
+  "files": [
+    "lib",
+    "index.js"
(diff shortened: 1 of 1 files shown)


kievechua:
@efossier Hi can you have a look a this pull request?

efossier:
Seems ok to me. +1

PR #236, opened 2017-04-19
add back deprecated ConversationsGrant

vassjozsef:


vassjozsef:
Talked to @markandrus about it, can we remove node 0.10?

tmconnors:
@vassjozsef Do not want to support node 0.10?

vassjozsef:
I do not think we should, it is more than two years old.

PR #241, opened 2017-05-02
Fix client.request to accept absolute url option

emckay:
I found a small bug in the `Client.prototype.request` function that caused the `if` statement to always resolve to true. This means that if an absolute URL is passed in `options.url`, the base url is added.


jingming:

PR #243, opened 2017-05-06
Support surrogate characters in xml builder, e.g. emoji’s

shawandrew:
Attempting to send emoji characters from inside a twiml message e.g. `twiml.message('😀');` will fail in the xml builder code due to a validation check against surrogate characters.  In order to make these supported we can edit the `MessagingResponse.js` to create the builder with support for these characters.

### DIFF
--- a/lib/twiml/MessagingResponse.js
+++ b/lib/twiml/MessagingResponse.js
@@ -8,7 +8,7 @@ var builder = require('xmlbuilder');
  * @constructor
  */
 function MessagingResponse() {
-  this.response = builder.create('Response').dec('1.0', 'UTF-8');
+  this.response = builder.create('Response', { allowSurrogateChars: true }).dec('1.0', 'UTF-8');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


efossier:
@shawandrew Thanks for the PR, looks good to me.

Can you please add a test which encodes the emoji in `spec/unit/twiml/MessagingResponse.spec.js`? I would do it myself but I cannot push to your fork. Once you do that I'll be happy to merge.

jingming:
@shawandrew If possible, do you mind adding this to `VoiceResponse` as well? If not, just the test should be fine and we will take care of rest!

shawandrew:
Updated with the test for the message response.  Sorry, I've not had experience of the voice response class yet so I'll leave that to you guys.

dbachko:
@shawandrew `allowSurrogateChars` option was removed in a current version of `xmlbuilder`. Now it is following XML 1.1 spec for valid chars instead of 1.0 which supports unicode chars.

shawandrew:
Great thanks!

PR #246, opened 2017-05-09
Add URL generation functions to TaskRouter JWTs

jingming:


lib/jwt/taskrouter/util.js
```diff
@@ -65,7 +66,99 @@ function defaultEventBridgePolicies(accountSid, channelId) {
   ];
 }
 
+/**
+ * Generate TaskRouter workspace url
+ *
+ * @param {string} [workspaceSid] workspace sid or '**' for all workspaces
```
efossier: maybe just default it to this if param is undefined so that people can just call `workspacesUrl()`?

efossier:
Seems reasonable

lib/jwt/taskrouter/util.js
jingming: you should be able to do that here. Is this what you are referring to? https://github.com/twilio/twilio-node/pull/246/files#diff-cce12a1e86ded2760558610cfe5ef0bcR16

efossier: Ah it looks like that will produce different behavior though, I guess thats what we want though. carry on.

PR #256, opened 2017-06-02
Add Node.js version to User-Agent header

ilanbiala:


jingming:
Looks fine, but this class is autogenerated.

PR #258, opened 2017-06-05
Test node 0.12 and latest stable version

ilanbiala:


.travis.yml
```diff
@@ -2,21 +2,16 @@ language: node_js
 sudo: false
 dist: precise
 node_js:
+    - "node"
```
efossier: is this how we specify node 8? Would prefer actually putting "8" if thats allowed or something that indicates that we are testing v8.

.travis.yml
```diff
@@ -2,21 +2,16 @@ language: node_js
...
     - "7"
     - "6"
     - "5"
     - "4"
     - "0.12"
-before_install:
-- npm install npm@2 -g
-- npm install npm -g
```
efossier: Not sure why these had to be added but the commits are from 2years ago so if everything works on travis... its probably fine?

ilanbiala: Yeah I looked at that, I think this should be fine.

.travis.yml
ilanbiala: This specifies simply the latest stable version of Node, which in this case is v8.0.0. I could add 8 explicitly and leave this in case the file isn't updated when a release comes out?

ilanbiala: Added Node 8 explicitly.

efossier: Yeah I think thats a good idea, we can do `- "8"` and also leave `- "node"` so we automatically test against latest as they update.

PR #259, opened 2017-06-08
Adding sim method for Sim noun in Dial

james-srl:
Dial is missing this method currently, probably because it is in beta.


james-srl:
Just realized this was a duplicate. My bad.

PR #261, opened 2017-06-13
Bug: POST requests being sent with wrong Content-Type.

djanowski:
POST requests are multipart, so setting it to `application/x-www-form-urlencoded` is wrong.

Request library takes care of setting the header to the right value.

### DIFF
--- a/lib/rest/Twilio.js
+++ b/lib/rest/Twilio.js
@@ -156,10 +156,6 @@ Twilio.prototype.request = function request(opts) {
   headers['User-Agent'] = 'twilio-node/' + moduleInfo.version;
   headers['Accept-Charset'] = 'utf-8';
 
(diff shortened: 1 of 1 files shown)


djanowski:
@efossier Can you please take a look? Twilio is probably ignoring the header, but it makes testing much harder. Thanks.

dougblack:
@djanowski Where have you seeing POST requests to Twilio needing to be of type `multipart/form-data`? `application/x-www-form-urlencoded` is the required content type for almost all APIs.

Still, letting the requests library take care of it seems like a good idea regardless. We will investigate. Can you elaborate on how `application/x-www-form-urlencoded` makes testing more difficult?

dougblack:
Closing due to inactivity.

PR #273, opened 2017-07-11
Add install requirement for Node

ilanbiala:


Makefile
```diff
@@ -1,4 +1,5 @@
 test-install:
+	@node --version || (echo "Node is not installed, please install Node >= 0.12"; exit 1);
```
efossier: probably can go in the `install` target since npm is unlikely to be installed if the user hasnt installed node and also it doesnt really make sense to install this package without node.

PR #279, opened 2017-07-24
Adds ChatGrant to jwt. Deprecates IpMessagingGrant.

philnash:
As the name IpMessaging has been deprecated, this changes the grant to the name `ChatGrant`. It is the same as the existing `IpMessagingGrant` so is just a straight name swap for developers.

`IpMessagingGrant` will continue to work and now inherits from `ChatGrant`. It will spit out a deprecation notice though.

I also silenced deprecations in test output as part of this.

### DIFF
--- a/lib/jwt/AccessToken.js
+++ b/lib/jwt/AccessToken.js
@@ -41,7 +41,7 @@ _.extend(TaskRouterGrant.prototype, {
  *                 assigned to the user
  * @param {string} options.pushCredentialSid - The Push Credentials SID
  */
-function IpMessagingGrant(options) {
+function ChatGrant(options) {
   options = options || {};
   this.serviceSid = options.serviceSid;
   this.endpointId = options.endpointId;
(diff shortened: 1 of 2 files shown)


lib/jwt/AccessToken.js
```diff
@@ -41,15 +41,15 @@ _.extend(TaskRouterGrant.prototype, {
...
   options = options || {};
   this.serviceSid = options.serviceSid;
   this.endpointId = options.endpointId;
   this.deploymentRoleSid = options.deploymentRoleSid;
   this.pushCredentialSid = options.pushCredentialSid;
 }
 
-_.extend(IpMessagingGrant.prototype, {
+_.extend(ChatGrant.prototype, {
   key: 'ip_messaging',
```
jingming: Should this key be `chat` or are we still using `ip_messaging` here?

philnash: That is a top question. Since the IpMessaging Grant still works I assumed it was because of the content of the token, including the key `ip_messaging`. I will check with the chat team to see if a `chat` key will work too.

philnash:
Good news @jingming, the Chat team confirmed that "chat" works as a key. I've updated accordingly.

philnash:
:tada:

PR #285, opened 2017-08-30
Autogenerate TwiML Classes

jingming:


lib/twiml/FaxResponse.js
```diff
@@ -0,0 +1,48 @@
...
+var builder = require('xmlbuilder');  /* jshint ignore:line */
+
+
+/* jshint ignore:start */
+/**
+ * <Response> TwiML for Faxes
+ */
+/* jshint ignore:end */
+function FaxResponse() {
+  this.response= builder.create('Response').dec('1.0', 'UTF-8');
```
efossier: nit: spacing.

lib/twiml/MessagingResponse.js
```diff
@@ -1,78 +1,102 @@
...
+ * <Message> TwiML Verb
  *
  * @param {object} attributes - ...
- * @param {string} [attributes.to] number to send message to
- * @param {string} [attributes.from] number to send from
- * @param {string} [attributes.method] (GET, POST)
- * @param {string} [attributes.action] action url
- * @param {string} [attributes.statusCallback] status callback url
- * @param {string} body body of message
+ * @param {phone_number} [attributes.to] Phone Number to send Message to
```
efossier: since we are now telling people to use the actual types instead of just `string`, will the xml library handle serializing these types to strings in the xml?

jingming: we don't have any types for these in node. these serve more as content hints even though you are just passing in a string.

PR #290, opened 2017-09-21
Node Security Project - Add nsp check script to warn for dependency vulnerabilities

bassettsj:
Why?
To verify that there are not known security vulnerabilities in our dependencies so we can actively address them when found.

Changes:
- add nsp as dev dependency (jshint was moved to be alphabetical by npm
cli)
- add nsp script to package.json
- add nsp to the make test command

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,7 @@ node_js:
     - "5"
     - "4"
     - "0.12"
+script: npm run ci
 notifications:
   slack:
     on_success: change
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -7,6 +7,7 @@ install:
 
 test:
 	npm test
+	npm run nsp
```
bassettsj: another option would be changing the `.travis.yml` to include this and linting as well.

dougblack:
Thanks @bassettsj, this should go out with our next release (likely tomorrow).

bassettsj:
@dougblack Hold on real quick, this should actually fail in CI. Let me move the script to the travis config and add a note in the contributing guide.

Issue #306, opened 2017-12-04
Validating Requests With Query String


chadwhitaker:
It appears that the `validateExpressRequest` function is no longer validating endpoints that contain a query param in the URL. Also, GET requests cannot be validated regardless of a query param.

Ref: https://github.com/twilio/twilio-node/issues/145

chadwhitaker:
cc: @jmctwilio

alexcchan:
@chadwhitaker and I touched base on this.  It looks like the root cause of validation failures is because `url.format` expects the `pathname` option to exclude query string, whereas Express `request.originalUrl` includes query string.

When `request.originalUrl` includes query string, the `?` separator gets URL-encoded by `url.format` as `%3F`, breaking the validation.

https://github.com/twilio/twilio-node/blob/master/lib/webhooks/webhooks.js#L50
https://nodejs.org/api/url.html#url_url_format_urlobject
http://expressjs.com/en/api.html#req.originalUrl

#311 is one way to work around this issue, but may not be the best way.

chadwhitaker:
Fixed via https://github.com/twilio/twilio-node/pull/311

PR #311, opened 2017-12-12
Fix X-Twilio-Signature validation when URL has ?

alexcchan:
`url.format` (https://nodejs.org/api/url.html#url_url_format_urlobject) treats the `pathname` option as the path _excluding_ query string, so any `?` get encoded as `%3F'.

Since (express) `request.originalUrl` may contain the query string, it is incorrect to pass it to `url.format` as a `pathname`.

This PR may not be the best solution, but is submitted to illustrate what the issue is.

When the `request.originalUrl` contains a `?` it should _not_ be URL-encoded as `%3F`.

https://github.com/twilio/twilio-node/issues/306

### DIFF
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -49,6 +49,9 @@ function validateExpressRequest(request, authToken, opts) {
         host: host,
         pathname: request.originalUrl
     });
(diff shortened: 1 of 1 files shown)


chadwhitaker:
Any thoughts on this being merged?

Issue #321, opened 2018-02-19
Validation fails with Proxy(HTTPS -> HTTP)


sakazuki:
**Version:**
2.11.1

### Issue
In the following case, `webhook(AUTH_TOKEN)` in **MyApp** cannot validate correctly.
**Twilio** ---(https)---> **Ngrok** ---(http)---> **MyApp**

The cause is that `webhook(AUTH_TOKEN)` uses "http" instead of "https" interally.

### Code Snippet
```node.js
(snip..)
app.post('/voice', twilio.webhook(AUTH_TOKEN), (req, res) => {
(snip..)
```

### Exception/Log
Voice logs in Twilio console
```
An attempt to retrieve content from https://0caxxxxx.ngrok.io/twilio returned the HTTP status code 403. 
```

### Steps to Reproduce
1.  create a test file as spec/test-proxy-ssl.js
```node.js
//const twilio = require('twilio');
const twilio = require('../lib/index');
const app = require('express')();
const bodyParser = require('body-parser');
const VoiceResponse = twilio.twiml.VoiceResponse;
const AUTH_TOKEN = process.env.TWILIO_AUTH_TOKEN;
const ngrok = require('ngrok');

app.use(bodyParser.urlencoded({extended: false}));

app.post('/voice', twilio.webhook(AUTH_TOKEN), (req, res) => {
  const response = new VoiceResponse();
  response.say(
    `Thanks for calling!
     Your phone number is ${req.body.From}. I got your call because of Twilio´s
     webhook. Goodbye!`
  );
  res.set('Content-Type', 'text/xml');
  res.send(response.toString());
});

app.listen(3000);

ngrok.connect(3000, function (err, url) {
  console.log("Set '" + url + "/voice' in Twilio VoiceIn webhook");
}); 
```
2. `npm install ngrok`
3. `TWILIO_AUTH_TOKEN=xxxxxxxxxxxxxxx node spec/test-proxy-ssl.js`
4. Set the output url in Twilio VoiceIn hook in Twilio console.
5. Call the phonenumber and get error

### Feature Request
* Please check **X-Forwarded-Proto** header to determine the protocol in `validateExpressRequest` of webhook.js

chadwhitaker:
Have you tried adding `app.enable('trust proxy');`?

sakazuki:
Hello @chadwhitaker ,

Thank you for your comment.
I test adding `app.enable('trust proxy');`, and I success to validate requests with proxy.
[...]

PR #322, opened 2018-02-19
feature: check X-Forwarded-Proto header to determine protocol

sakazuki:
perform the #321  request.


sakazuki:
I cancel this request. Because  `app.enable('trust proxy');` is better solution.
See detail #311

PR #324, opened 2018-02-26
Update version of hoek to 4.2.1

jmctwilio:


jmctwilio:
Looks like dropping future support for node v0.12 is the best course of action.

Issue #295, opened 2017-10-16
TypeError: Converting circular structure to JSON


subodhpareek18:
**Version:**
3.8.1

### Code Snippet
```js
const twilio = new Twilio(creds.accountSid, creds.authToken);
const chatClient = twilio.chat.services(creds.chatServiceSid);

// ....
JSON.stringify(chatClient.channels(id).fetch())
```

### Exception/Log
```
TypeError: Converting circular structure to JSON
```

subodhpareek18:
For now I've written this utility that removes some of the keys (that I anyways didn't really want) and that removes all the circular references.

I think there should be an options object that we could pass while instantiating twilio which could remove these references for us.

```js
import R from 'ramda';

const dangerous = ['_version', '_context', '_solution'];
const omit = R.omit(dangerous);

export default data => {
  if (R.is(Array, data)) {
    return { data: R.map(omit, data) };
  } else {
    return omit(data);
  }
};
```

efossier:
@subodhpareek18 This worked for me, perhaps different versions of node fail to deserialize the data? Which version are you running? 

```
> client.chat.services("IS1234").channels("CH1234").fetch().then((res) => console.log(res));
{ state: 'pending' }
> ChannelInstance {
  _version:
   V2 {
     _domain:
      Chat {
        twilio: [Object],
        baseUrl: 'https://chat.twilio.com',
        _v1: undefined,
        _v2: [Circular] },
     _version: 'v2',
... } }
```

subodhpareek18:
Sorry, I wasn't fully clear on my requirements, so when I `console.log` the above then it gives me this Object `ChannelInstance` and that's fine, but I need to send the response back to the client that called my method. 

For that to happen I need to send it as JSON, because only JSON can travel across HTTP not JS Objects.

```js
const channel = await this.client.channels(id).fetch()
return Promise.resolve(channel)
```
This is where I get an error where my web server framework tries to JSON.stringify the channel object.

subodhpareek18:
I think a `{ lean: true }` or `{ raw: false }` would be a good idea. Many ORMs like `sequelize` and `mongoose` do it. Or else there could be a `.toJSON()` method.

subodhpareek18:
This will also improve the performance, especially for find calls. If the user simply wants a plain object without any context, etc then the helper library could circumvent Object Creation and simply return the response it got from the twilio api.

efossier:
[...]

PR #325, opened 2018-02-27
Example generating toJSON on MessageInstance

efossier:
The actual change was made in our code generation tool, this is just an example on the `MessageInstance` class, once the code-generation side is merged this method will be available on every instance class.


efossier:
Fix for https://github.com/twilio/twilio-node/issues/295

cymen:
This looks great to me -- I've been using the workaround of filtering out `_version`, `_context` and `_solution` but I only applied that approach as a temporary workaround.

efossier:
This will be included in next release

PR #326, opened 2018-02-28
Add ability to mock twillio

APshenkin:
As a QA, I want to be able, to change twillio url to forward requests to mock server to test different responses from twillio. This change will give access to set twillio base url to mock server

### DIFF
--- a/lib/rest/Api.js
+++ b/lib/rest/Api.js
@@ -67,7 +67,7 @@ var V2010 = require('./api/V2010');  /* jshint ignore:line */
  */
 /* jshint ignore:end */
 function Api(twilio) {
-  Domain.prototype.constructor.call(this, twilio, 'https://api.twilio.com');
+  Domain.prototype.constructor.call(this, twilio, twillio.baseUrl);
 
   // Versions
   this._v2010 = undefined;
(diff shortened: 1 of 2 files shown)


efossier:
@APshenkin The files you changed are generated code, we can investigate supporting this in our code generation tool but it may be a while until we get to it as our team is swamped with work.

dkundel:
For the thing you want to do there is really no code change on our side necessary from what I see. While your proposed change could be a solution, this might just do the job for you:

You can pass to the options of the [Client](https://github.com/twilio/twilio-node/blob/master/lib/rest/Twilio.js) a custom HttpClient. It has to fulfill the same interface as `RequestClient` in [RequestClient.js](https://github.com/twilio/twilio-node/blob/master/lib/base/RequestClient.js) but then you could simply do a dynamic replacement of the respective URLs with whatever host you want to place it.

Example: 

```js
const twilio = require('twilio');
const RequestClient = require('twilio/lib/base/RequestClient');

class MyMockRequestClient {
  constructor() { this.http = new RequestClient(); }
  request(opts) {
     opts.uri = opts.uri.replace('https://api.twilio.com', 'https://my-fake-api.example.com'); 
     return this.http.request(opts);
  }
}

const client = twilio('accountSid', 'authToken', { httpClient: new MyMockRequestClient() });
```

This example simply overrides the URI and then uses the base HTTP `RequestClient` that the Twilio lib uses. But obviously you could also do completely different things.

Cheers,
Dominik

APshenkin:
@dkundel Thanks! It works

PR #355, opened 2018-06-13
Add addText method to TwiML classes

ekarson:
Allow adding arbitrary text under TwiML nodes via the new method `addText`

### DIFF
--- a/lib/twiml/MessagingResponse.js
+++ b/lib/twiml/MessagingResponse.js
@@ -47,10 +47,12 @@ MessagingResponse.prototype.message = function message(attributes, body) {
  * @param {object} attributes - ...
  * @param {string} [attributes.method] Redirect URL method
  * @param {url} url Redirect URL
+ *
+ * @returns Redirect
  */
 /* jshint ignore:end */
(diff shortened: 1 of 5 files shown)


lib/twiml/MessagingResponse.js
```diff
@@ -47,10 +47,12 @@ MessagingResponse.prototype.message = function message(attributes, body) {
  * @param {object} attributes - ...
  * @param {string} [attributes.method] Redirect URL method
  * @param {url} url Redirect URL
+ *
+ * @returns Redirect
  */
 /* jshint ignore:end */
 MessagingResponse.prototype.redirect = function redirect(attributes, url) {
-  this.response.ele('Redirect', attributes, url);
+  return new Redirect(this.response.ele('Redirect', attributes, url));
```
cjcodes: Can you explain what the change from `response.ele` -> `new Element` does here? Was `response.ele` just the object before?

ekarson: `response.ele` means "add a new element with the name 'Redirect'" -- it mutated the `response` without returning anything.

The new code still does the mutation, but then it returns the newly-created element node (wrapped in a class named after the element).

PR #356, opened 2018-06-14
Update nsp to latest

ekarson:
Updated the version of nsp in `package.json` and regenerated `package-lock.json`


cjcodes:
Actually--can we add a note to the README.md which versions of node we support? And maybe include a `>=8` in package.json? (I think you can do this, right?)

ekarson:
@cjcodes It should be `>=6`, right? (Updated)

PR #357, opened 2018-06-15
Allow creating generic twiml nodes

cjcodes:
- Creates a basic class inheritance between `Response`s and `TwiML` (added)
- Moves `addText()` method to parent `TwiML` class
- Adds a `addChild()` method to anything that inherits `TwiML`

### DIFF
--- a/lib/twiml/TwiML.js
+++ b/lib/twiml/TwiML.js
@@ -1,0 +1,86 @@
+'use strict';
+
+var builder = require('xmlbuilder');  /* jshint ignore:line */
+
+/* jshint ignore:start */
+/**
+ * Parent TwiML object
+ */
+/* jshint ignore:end */
+function TwiML() {
+  this.response = builder.create('Response').dec('1.0', 'UTF-8');
+}
+
+/* jshint ignore:start */
+/**
+ * Because child elements have properties named after their class names, we need
+ * to translate that when we want to access that at the parent prototype level.
+ * So this will translate something like Say to 'say' and VoiceResponse to
+ * 'response'.
+ *
+ * It will also soft-cache it on the object so we don't end up doing string
+ * comparisons. Save a few CPU cycles.
+ */
+/* jshint ignore:end */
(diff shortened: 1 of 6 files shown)


lib/twiml/TwiML.js
```diff
@@ -0,0 +1,86 @@
...
+
+/* jshint ignore:start */
+/**
+ * Convert to TwiML
+ *
+ * @returns TwiML XML
+ */
+/* jshint ignore:end */
+TwiML.prototype.toString = function toString() {
+  return this[this._getPropertyName()].end();
```
ekarson: Is there a way to have a non-public `getXml` that returns `this[this._getPropertyName()]` to avoid duplication?

cjcodes: Duplication of what?

ekarson: the code `this[this._getPropertyName()]` appears in several places -- I thought it would be nice to have a helper method to avoid repeating it

lib/twiml/TwiML.js
```diff
@@ -0,0 +1,74 @@
...
+/* jshint ignore:start */
+/**
+ * Because child elements have properties named after their class names, we need
+ * to translate that when we want to access that at the parent prototype level.
+ * So this will translate something like Say to 'say' and VoiceResponse to
+ * 'response'.
+ */
+/* jshint ignore:end */
+TwiML.prototype._getPropertyName = function _getPropertyName() {
+  return this._propertyName === undefined ? 'response' : this._propertyName;
```
ekarson: is it hard to get the response classes to emit a `propertyName` as well? this seems like kind of a weird special case

ekarson: we can remove this logic now, right?

lib/twiml/TwiML.js
cjcodes: Ah, yes, great point!

lib/twiml/TwiML.js
cjcodes: Done.

PR #361, opened 2018-06-29
Regenerate with structured params/properties

ekarson:
Regenerated with updates to yoyodyne that store structured param/property/docstring data (rather than keeping everything in strings).

Docstrings now all match function param lists.


ekarson:
@dkundel Here's the output from running my yoyodyne branch -- could you take a look and let me know if it seems reasonable? (This is just hopefully cosmetic changes to the Javascript source -- I will open another PR to restore the Typescript declarations.)

lib/rest/accounts/v1/credential/aws.js
```diff
@@ -66,34 +66,35 @@ AwsList = function AwsList(version) {
    * @memberof Twilio.Accounts.V1.CredentialContext.AwsList
    * @instance
    *
-   * @param {object|function} opts - ...
+   * @param {object} [opts] - ...
```
dkundel: Why are you removing `function` as a valid type for `opts` here. The argument can be a callback function as well if you are not planning to pass in options

lib/rest/accounts/v1/credential/aws.js
```diff
@@ -66,34 +66,35 @@ AwsList = function AwsList(version) {
...
    */
   /* jshint ignore:end */
   AwsListInstance.each = function each(opts, callback) {
-    opts = opts || {};
     if (_.isFunction(opts)) {
-      opts = { callback: opts };
-    } else if (_.isFunction(callback) && !_.isFunction(opts.callback)) {
-      opts.callback = callback;
+      callback = opts;
+      opts = {};
```
dkundel: This logic seems flawed. I didn't try it but in my head if I run this:

```js
function someCallback(item) { console.log(item) }
each({ pageSize: 20 }, someCallback);
```

Then `opts` would be overridden in line 91 and therefore my options being dropped

lib/rest/api/v2010/account/address.js
```diff
@@ -504,7 +499,8 @@ AddressPage.prototype.getInstance = function getInstance(payload) {
  *
  * @param {Twilio.Api.V2010} version - Version of the resource
  * @param {object} payload - The instance payload
- * @param {sid} accountSid - The account_sid
+ * @param {sid} accountSid -
```
dkundel: Ideally these types such as 'sid' should be typed as `string` in JSDoc comments. Same goes for the other ones. It's cool if we create custom type aliases like `type Sid = string` for them in TypeScript (that's what I did with some of them) but I think for the JSDoc comments actually naming them as strings and then having in the description that they have to have the shape of a SID (or a specific date style) seems more reasonable

lib/rest/accounts/v1/credential/aws.js
ekarson: `opts` is marked as optional, so conceptually, you can either provide an `object` for `opts` and a function for `callback`, or omit `opts` and provide a `function` for `callback`. Either way, the `callback` is a `function` and `opts` is an `object`.

(Of course, internally, we need to swap them around, but that does not seem like a relevant detail for the consumer.)

lib/rest/accounts/v1/credential/aws.js
ekarson: no, line 91 only executes if you omit the `opts` argument. I'll update the test case to include this.

lib/rest/api/v2010/account/address.js
ekarson: Since the type is already `sid` and this change is so massive, I would rather leave fixing this for another time.

ekarson:
Thanks for reviewing @dkundel -- I added tests to cover the case you mentioned.

The whole idea of being able to have an omitted argument before a present argument seems insane to me, and I would just as soon not encourage clients to take advantage of it. If there's a generally accepted best practice to indicate optional parameters with `{type1|type2}` notation, then I'm happy to revisit it. I couldn't find any examples (all my searches just turned up articles saying "don't do this in the first place").

PR #363, opened 2018-07-06
Rough draft of TS declarations for options interfaces in Twilio client

ekarson:
Testing the pipeline for generating TS declarations on the Twilio client class. The output is not correct yet, just showing that we can once again generate them at all.


lib/rest/Twilio.js
```diff
@@ -550,4 +549,11 @@ Object.defineProperty(Twilio.prototype,
   }
 });
 
-module.exports = Twilio;
+module.exports = {
+  T: T,
```
dkundel: What broke here? This doesn't seem right

ekarson:
whoops -- thanks @dkundel! fixed

PR #365, opened 2018-07-10
Add a request body validator

cjcodes:
This PR:

- Adds a request body validator that accepts a string, for validating POST body that's not form data
- Gracefully (backwards-compatibly) removes unsupported, useless code from the Express Middleware
- Refactors tests to not stand up an express server, mocking requests instead

### DIFF
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -18,7 +18,35 @@ function validateRequest(authToken, twilioHeader, url, params) {
       url = url + key + params[key];
   });
 
-  return scmp(twilioHeader, crypto.createHmac('sha1', authToken).update(new Buffer(url, 'utf-8')).digest('Base64'));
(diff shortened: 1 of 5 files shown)


lib/webhooks/webhooks.js
```diff
@@ -52,14 +80,24 @@ function validateExpressRequest(request, authToken, opts) {
...
+      request.header('X-Twilio-Signature'),
+      webhookUrl,
+      request.body || {}
+    );
+  } else {
+    return validateRequest(
+      authToken,
+      request.header('X-Twilio-Signature'),
+      webhookUrl,
+      request.body || {}
```
ekarson: wait does `request.body` contain the params?

cjcodes: Yup, same name. In express, `request.body` is auto-converted based on the content-type.

I think that's your question, right?

PR #382, opened 2018-08-31
Add method overloading and resolve array inputs

cjcodes:
Regenerate with resolved TS definitions.


dkundel:
I don't think we need this one anymore, so I'm closing this one.

PR #385, opened 2018-09-13
Include TTL option in ClientCapability

daviddelucca:
Included TTL option in `ClientCapability` and test code to cover this.

I was not sure about write another condition in unit test where a token is generated so I wrote a new unit test to cover TTL option. Let me know If I should change it and move the condition to another test.


dkundel:
Thanks @daviddelucca! This looks good to me. @cjcodes can you briefly take a look at this and merge it in when you get a chance?

PR #391, opened 2018-10-15
DeprecationWarning: Buffer()

ShelbyZ:
After running `npm test` the log mentions a deprecation warning:

(node:12636) [DEP0005] DeprecationWarning: Buffer() is deprecated due to security and usability issues. Please use the Buffer.alloc(), Buffer.allocUnsause the Buffer.alloc(), Buffer.allocUnsafe(), or Buffer.from() methods instead.

This fix should be in line with supporting Node 6/8 and other areas of the code are already using Buffer.from(...).

### DIFF
--- a/lib/base/RequestClient.js
+++ b/lib/base/RequestClient.js
@@ -40,7 +40,7 @@ RequestClient.prototype.request = function(opts) {
   }
 
   if (opts.username && opts.password) {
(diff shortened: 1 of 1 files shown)


dkundel:
Since we don't support anything < Node.js 6.0 anymore this should be fine to merge. @cjcodes can you merge this in?

PR #399, opened 2018-11-30
Fixed notification resource types

JavanPoirier:
Changed defined types according to notification resource:
https://www.twilio.com/docs/notify/api/notification-resource

*Note: first time making a public pull request. I will greatly appreciate your feedback and please let me know if I did something wrong. I will gladly fix it, I'm here to learn! Thanks

### DIFF
--- a/lib/rest/notify/v1/service/notification.d.ts
+++ b/lib/rest/notify/v1/service/notification.d.ts
@@ -86,19 +86,19 @@ interface NotificationResource {
   alexa: string;
   apn: object;
   body: string;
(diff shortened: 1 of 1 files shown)


childish-sambino:
Looks like we're typing objects as strings everywhere. This affects other resources and is part of a generated file. I'll add a task to the backlog to fix them all. Closing since this is in generated code and would be overwritten on future releases.

childish-sambino:
Fixed by https://github.com/twilio/twilio-node/pull/475

PR #404, opened 2018-12-14
[BUG] Fix/use lts node versions and remove eol versions

euxn23:
Node versions used in travis `0.10` and `0.8` is already EOL.
Remove then and set LTS versions.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
 language: node_js
 sudo: false
 node_js:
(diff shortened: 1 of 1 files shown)


kamronbatman:
Would it make sense to make this more generic to the supported LTS versions?
`"10"` instead of `"10.14.2"` since the newest version is no longer that one?

childish-sambino:
Closing in favor of #436

PR #414, opened 2019-02-14
Fix return type for remove()

cdock1029:
getting tslint error:

```
- Expression has type `void`. Put it on its own line as a statement.
```
when writing this code:

```ts
await twilioClient.incomingPhoneNumbers(sid).remove()
```


childish-sambino:
The return type should really be `Promise<boolean>` (promise that resolves to true if record was deleted) and it should apply to all `remove` methods. I'll add a task to the backlog to address this since this is a generated file and future releases would just wipe the change away.

PR #417, opened 2019-02-25
Remove the auto-generated docs from source control

childish-sambino:
These docs are auto-generated during release and available here: https://www.twilio.com/docs/libraries/reference/twilio-node/

They do **not** need to reside under source control.


cjcodes:
My only open question is if these are powering something we don't know about. We don't have GitHub pages, but is this powering something like readthedocs.com or something we don't know about?

PR #421, opened 2019-02-27
add typedef for validateRequestWithBody

### LINKED ISSUE #420: `validateRequestWithBody` is missing type def
**Version: 3.28.1**

[`validateRequestWithBody`](https://github.com/twilio/twilio-node/blob/master/lib/webhooks/webhooks.js#L38) is missing a type definition.

Charliekenney23:
Closes #420.


Charliekenney23:
@childish-sambino thanks for the quick feedback!

PR #437, opened 2019-04-02
Fix parsing error with lodash

travoltron:
When using this library with Babel and lodash, an error arises when trying to send an SMS, or make any changes that require the _.template function to get called. 
`SyntaxError: Unexpected token = at Function (<anonymous>)`, specifically `source: 'function(obj) {\nobj || (obj = {});\nvar __t, __p = \'\', __j = Array.prototype.join;\nfunction print() { __p += __j.call(arguments, \'\') }\nwith (obj) {\n__p += \'/Accounts/\';\n= sid ;\n__p += \'.json\';\n\n}\nreturn __p\n}'`

To alleviate this, I broke up the template parsing to use concatenated strings, rather than interpolation. 

**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [X] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/lib/rest/api/v2010/account.js
+++ b/lib/rest/api/v2010/account.js
@@ -992,7 +992,7 @@ AccountContext = function AccountContext(version, sid) {
   // Path Solution
   this._solution = {sid: sid, };
   this._uri = _.template(
-    '/Accounts/<%= sid %>.json' // jshint ignore:line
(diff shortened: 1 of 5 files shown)


childish-sambino:
@travoltron Could you provide sample code that illustrates the issue you're running into?

PR #447, opened 2019-05-07
Check X-Twilio-Signature in webhook middleware

neerajwadhwa:
Implemented a check in lib/webhooks/webhooks.js's webhook() middleware to check if the X-Twilio-Signature header exists or not.

If the request is not made from Twilio servers then we won't have *X-Twilio-Signature* header in the request and the middleware will throw cryptic error from *validateRequest()* function.

### DIFF
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -171,6 +171,12 @@ function webhook() {
 
   // Create middleware function
   return function hook(request, response, next) {
+    // Check if the 'X-Twilio-Signature' header exists or not
+    if (!request.header('X-Twilio-Signature')) {
+      return response.type('text/plain')
+      .status(400)
+      .send('No signature header error - X-Twilio-Signature header does not exist, maybe this request is not coming from Twilio.');
+    }
     // Do validation if requested
     if (opts.validate) {
       // Check for a valid auth token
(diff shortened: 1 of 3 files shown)


khalilchoudhry:
@neeraj-wadhwa-cc I was wondering, why [this check](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L176) is added here before [checking](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L182) the validate option. I am sending requests from POSTMAN and I have not set X-Twilio-Signature. I am on dev/test environment, which means the validate option is false, so there is no need to check if the header is present or not. Similarly if I have written test for Express routes in my code, the middleware will fail despite the fact that validate option is false.

khalilchoudhry:
It also checks for X-Twilio-Signature [here](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L50) which will not fail with validate set to false.

childish-sambino:
@khalilchoudhry The changes mentioned were made in very close time proximity so the coordination wasn't perfect. Will follow-up on #460

PR #446, opened 2019-05-07
Assign default value to twilioHeader during validation

mniehe:
Adds a sensible default to `twilioHeader` if it's undefined during validation.

Closes #445 

**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #460, opened 2019-06-21
Request middleware fails on validate set to false

khalilchoudhry:
[X-Twilio-Signature check](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L176) is added before [checking](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L182) the validate option. The request will fail even if [validate option](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.d.ts#L32) is set to false. If a request is sent from POSTMAN on dev/test environment, which means the validate option is false, so there is no need to check if the header is present or not. Similarly, if a test is written for Express route in my code, the middleware will fail despite the fact that validate option is false.

It also checks for X-Twilio-Signature [here](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L50) which will not fail with validate option set to false. If validate option is true and X-Twilio-Signature is undefined, empty string is set as a [default value](https://github.com/twilio/twilio-node/pull/446) which will then obviously differ from [expected twilio signature](https://github.com/twilio/twilio-node/blob/6585f327e4f681971dc3eddf40ca5840adf685dc/lib/webhooks/webhooks.js#L51) and fail.
**Contributing to Twilio**

### DIFF
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -172,12 +172,6 @@ function webhook() {
 
   // Create middleware function
   return function hook(request, response, next) {
-    // Check if the 'X-Twilio-Signature' header exists or not
-    if (!request.header('X-Twilio-Signature')) {
-      return response.type('text/plain')
-      .status(400)
-      .send('No signature header error - X-Twilio-Signature header does not exist, maybe this request is not coming from Twilio.');
-    }
     // Do validation if requested
     if (opts.validate) {
       // Check for a valid auth token
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -174,6 +174,12 @@ function webhook() {
   return function hook(request, response, next) {
     // Do validation if requested
     if (opts.validate) {
+      // Check if the 'X-Twilio-Signature' header exists or not
+      if (!request.header('X-Twilio-Signature')) {
+        return response.type('text/plain')
+        .status(400)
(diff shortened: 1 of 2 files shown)


khalilchoudhry:
@childish-sambino Please review this.

childish-sambino:
Instead of removing the check from `webhook()`, how about just moving it down to somewhere within the validation block? This would then support the functionality you're looking for _and_ provide a more specific error message and proper response code. I prefer to be more specific in the message as long as it's not a security risk.

khalilchoudhry:
Isn't X-Twilio-Signature is expected from Twilio only? How can someone create or set X-Twilio-Signature on dev environment or during testing? Why is it asked [here in docs](https://www.twilio.com/docs/usage/tutorials/how-to-secure-your-express-app-by-validating-incoming-twilio-requests#disable-request-validation-during-testing) to disable validation during testing then ?

childish-sambino:
If you move the header check to within the `if (opts.validate) {` block, then it will not be checked during dev/test.

khalilchoudhry:
done

Issue #481, opened 2019-09-25
Unable to specify CA cert bundle to override request default


whitebarry:
I'm troubleshooting local connectivity issues when using this module with my corporate security and came across this feature of Node that allows users to specify additional CA certs for making HTTPS requests. However, the Twilio client does not seem to respect this setting as I get the same error with the variable set or not.

**Version:**
3.35.0

### Code Snippet
```node
const dotenv = require('dotenv');
dotenv.config();
console.log(`Using extra certs: ${process.env.NODE_EXTRA_CA_CERTS}`)

const Twilio = require('twilio');

const startDate = new Date();
startDate.setHours(0, 0, 0, 0);
const client = new Twilio.Twilio(process.env.ACCOUNT_SID, process.env.AUTH_TOKEN);
client.taskrouter
    .workspaces(process.env.WORKSPACE_SID)
    .statistics('')
    .fetch({ taskChannel: 'voice', startDate }).then((stats) => {
    console.log(stats);
}).catch((error) => {
    console.log(error);
});
```

### Exception/Log
```
Using extra certs: C:\path\to\my\cert.crt
{ Error: unable to get local issuer certificate
    at TLSSocket.onConnectSecure (_tls_wrap.js:1058:34)
    at TLSSocket.emit (events.js:198:13)
    at TLSSocket._finishInit (_tls_wrap.js:636:8) code: 'UNABLE_TO_GET_ISSUER_CERT_LOCALLY' }
```

### Steps to Reproduce
1. Be on a corporate network that requires an additional CA cert bundle
2. Execute the above code

PR #483, opened 2019-09-25
Added the ability for users to specify a CA bundle for HTTPS requests

whitebarry:
Proposal to Issue #481

Right now there is no way for a user to specify a CA bundle to replace the default one that `request` ships with.

According to [the request docs](https://github.com/request/request#tlsssl-protocol) you can specify it in the `options` object passed to `request`.

This change offers a simple solution by setting a path to your CA bundle as an environment variable: 
`TWILIO_CA_BUNDLE=/path/to/cert.pem`

Test cases have been included to ensure that the specified CA bundle is used, and that the original behavior of not including it in `options` remains unchanged if it is not specified.

**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/lib/base/RequestClient.js
+++ b/lib/base/RequestClient.js
@@ -55,7 +55,10 @@ RequestClient.prototype.request = function(opts) {
   };
 
   if (process.env.TWILIO_CA_BUNDLE !== undefined) {
-    options.ca = fs.readFileSync(process.env.TWILIO_CA_BUNDLE);
+    if (this.ca === undefined) {
+      this.ca = fs.readFileSync(process.env.TWILIO_CA_BUNDLE);
+    }
+    options.ca = this.ca;
   }
 
   if (!_.isNull(opts.data)) {
--- a/lib/base/RequestClient.js
+++ b/lib/base/RequestClient.js
@@ -2,6 +2,7 @@
 
 var _ = require('lodash');
 var http = require('request');
+var fs = require('fs');
 var Q = require('q');
 var Response = require('../http/response');
 var Request = require('../http/request');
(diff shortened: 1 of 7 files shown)


lib/base/RequestClient.js
```diff
@@ -53,6 +54,10 @@ RequestClient.prototype.request = function(opts) {
     forever: opts.forever === false ? false : true,
   };
 
+  if (process.env.TWILIO_CA_BUNDLE !== undefined) {
```
childish-sambino: Could you also add a block to the README about this new env var and the existing ones that are supported?

```
TWILIO_ACCOUNT_SID ...
TWILIO_AUTH_TOKEN ...
TWILIO_CA_BUNDLE ...
```

whitebarry:
>
>
> One request to update the docs

Can you point me to a specific place in the docs you'd like me to update? I'm not too familiar with this module yet.

> 
> 
> assuming you tested this locally and it resolves the issue you're having

Yes it works like a charm locally :smile:

childish-sambino:
@whitebarry I think sticking a block in/under here is fine: https://github.com/twilio/twilio-node#sample-usage

whitebarry:
Ok added some info to the example file and also the `README` block for environment variables. I'm not exactly clear on the usage of the first 2 you mentioned so I left the descriptions for those as TODOs, hope that's OK.

whitebarry:
> since the CA payload is added to every API request, it would be better to cache the file contents on the first load

I couldn't agree more. My latest change addresses this with a supporting test case.

childish-sambino:
FYI, next planned release is on 2019-10-02.

PR #484, opened 2019-10-07
Fixes missing identity value in token when it's not a string

### LINKED ISSUE #438: Missing identity in token when not string
*Note: These issues are for bugs and feature requests for the helper libraries.
If you need help or support, please email help@twilio.com and one of our experts
will assist you!*

**3.29**

### Code Snippet
```node
function getTwilioToken(some_id) {
  const token = new AccessToken(
    TWILIO_ACCOUNT_SID,
    TWILIO_API_KEY,
    TWILIO_API_SECRET, {
      identity: `${some_id}`
    }
  );
  const chatGrant = new ChatGrant({
    serviceSid: TWILIO_CHAT_SERVICE_SID
  });
  token.addGrant(chatGrant);
  return token;
}
```

### Steps to Reproduce
1. Send in an int

### Feature Request
We should handle/throw error, if the identity is integer or not string. Currently, it produces token but missing the identity field.

codedawi:
Fixes #438

Resolves missing identity value in token when it's not a string begin passed in the `AccessToken` constructor.

The specs of the issue ticket suggest throwing an error when identity is passed as an integer. Originally, I took this approach but I quickly realized this would cause breaking changes for anyone passing numeric values in identity option. I do not believe this is error worthy because simple converting the value to a string solves the down stream implication in the jwt.

```javascript
 if (_.isInteger(options.identity)) { options.identity = new String(options.identity); };
```

**Contributing to Twilio**

### DIFF
--- a/lib/jwt/AccessToken.js
+++ b/lib/jwt/AccessToken.js
@@ -249,7 +249,7 @@ _.extend(AccessToken.prototype, {
     }
 
     var grants = {};
-    if (_.isInteger(this.identity) || _.isString(this.identity)) { grants.identity = String(this.identity); };
+    if (_.isInteger(this.identity) || _.isString(this.identity)) { grants.identity = String(this.identity); }
 
     _.each(this.grants, function(grant) {
       grants[grant.key] = grant.toPayload();
--- a/lib/jwt/AccessToken.js
+++ b/lib/jwt/AccessToken.js
@@ -212,6 +212,7 @@ function AccessToken(accountSid, keySid, secret, options) {
   if (!keySid) { throw new Error('keySid is required'); }
   if (!secret) { throw new Error('secret is required'); }
   options = options || {};
+  if (_.isInteger(options.identity)) { options.identity = new String(options.identity); };
 
   this.accountSid = accountSid;
   this.keySid = keySid;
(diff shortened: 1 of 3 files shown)


lib/jwt/AccessToken.js
```diff
@@ -212,6 +212,7 @@ function AccessToken(accountSid, keySid, secret, options) {
   if (!keySid) { throw new Error('keySid is required'); }
   if (!secret) { throw new Error('secret is required'); }
   options = options || {};
+  if (_.isInteger(options.identity)) { options.identity = new String(options.identity); };
```
childish-sambino: I'd rather fix the check on line 252 below. Could it just be `grants.identity = String(this.identity);`?

codedawi:
@childish-sambino yes makes sense to move it there. I left the original conditional statement checking `_.isString` and added `_.isInteger` with an "or" operator. When true,  `grants.identity = String(this.identity);`

lib/jwt/AccessToken.js
```diff
@@ -249,7 +249,7 @@ _.extend(AccessToken.prototype, {
     }
 
     var grants = {};
-    if (_.isString(this.identity)) { grants.identity = this.identity; }
+    if (_.isInteger(this.identity) || _.isString(this.identity)) { grants.identity = String(this.identity); };
```
childish-sambino: nit: Trailing semicolon not needed.

PR #487, opened 2019-10-08
Allow for auto-deploy on tagged releases

thinkingserious:
**Contributing to Twilio**

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,23 +1,31 @@
 language: node_js
 node_js:
-    - "node"
-    - "lts/*"
-    - "8"
-    - "6"
+- "node"
+- "lts/*""
+- "8"
+- "6"
 services:
-  - docker
+- docker
 script: npm run ci
 notifications:
   slack:
     on_success: change
     on_failure: change
     rooms:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -15,9 +15,17 @@ notifications:
...
+    skip_cleanup: true
+    on:
+      tags: true
+      node: "8"
+  - provider: npm
+    email: twilio-ci@twilio.com
+    api_key:
+      secure: aseNt4M8/qLU5+Wh5f8Bqy8QEl6BT4id4qaq0fhRiVSrAOyb3em2nevjRRdZCVkLQvlV/v4uceo1h3kebXlj/4q9bHu0UnyNVnC9t+bZrSCRXfxWcvsq5XPKOwtXmLuVgNOhp6Zhd3Gaex4gyCummhMelXclLYhwOkp++SE0prY=
+    on:
+      repo: twilio/twilio-node
```
childish-sambino: Don't think the `repo` line is needed.

thinkingserious: Yes, it's just one extra level of security, I think. I'm OK with leaving it out.

Do you see a scenario where leaving it in would not be advisable?

childish-sambino: It's just more lines of configuration. If leaving it in, should add to the other deploy target too.

PR #491, opened 2019-10-18
added request validation for urls with and without ports

### LINKED ISSUE #411: validateExpressRequest as documented works with twiml but not event callbacks
Not sure if this is the right place - its not technically a problem with the twilio-node library, but with the documentation.

**Version:**
3.28.0
### Code Snippet
```node
let host = req.headers.host;
if (req.protocol === 'https') {
  // Strip off the port, like the requesting twilio server would have
  host = host.substring(0, host.indexOf(':'));
}
const valid = twilio.validateExpressRequest(req, authToken, {host: host});
```

### Steps to Reproduce
1. Have a server using non-standard https port (e.g. 3443)
2. Use the above snippet to process an incoming request from Twilio for TwiML
3. Use the above snippet to process an incoming request from Twilio that is a callback for event (e.g. conference callback)

In (2), it will work. In (3) it will not, unless you remove the line that strips the port.

This indicates that when Twilio generates the signature it strips the port when preparing a request for TwiML, but not for a callback. The reason I was stripping out the port in the first place was because https://www.twilio.com/docs/usage/security states:

>For SMS and voice callbacks over HTTP, Twilio will drop the username and password (if any) from the URL before computing the signature. For SMS and voice callbacks over HTTPS, Twilio will drop the username, password and port (if any) before computing the signature. This behavior will continue to be supported in the 2008-08-01 and 2010-04-01 versions of the API to ensure compatibility with existing code. We understand this behavior is inconsistent, and apologize for the inconvenience.

I assumed in the context of the documentation "callback" means any request from twilio to another server in the process of calls (so twiml or reporting an event). However it seems to be different for different circumstances.

I tried finding other issues/discussions about this, but it was related to stripping out the _standard_ ports, not nonstandard ones.

eshanholtz:
PR for DI-320
Fixes #411

### DIFF
--- a/lib/webhooks/webhooks.js
+++ b/lib/webhooks/webhooks.js
@@ -3,7 +3,27 @@
 var crypto = require('crypto');
 var _ = require('lodash');
 var scmp = require('scmp');
-var url = require('url');
+var urllib = require('url');
+
+/**
+ * Utility function to construct the URL string, since Node.js url library won't include standard port numbers
+ *
+ * @param {URL} parsedURL - The parsed url object that Twilio requested on your server
+ * @returns {string} - URL with standard port number included
+ */
+function buildUrlWithStandardPort(parsedUrl) {
+  let url = '';
+  const port = parsedUrl.protocol === 'https:' ? ':443' : ':80';
+
+  url += parsedUrl.protocol ? parsedUrl.protocol + '//' : '';
+  url += parsedUrl.username;
+  url += parsedUrl.password ? ':' + parsedUrl.password : '';
+  url += (parsedUrl.username || parsedUrl.password) ? '@' : '';
+  url += parsedUrl.host ? parsedUrl.host + port : '';
+  url += parsedUrl.pathname + parsedUrl.search + parsedUrl.hash;
+
+  return url;
+}
 
 /**
  Utility function to add a port number to a URL
@@ -13,7 +33,7 @@ var url = require('url');
  */
 function addPort(parsedUrl) {
   if (!parsedUrl.port) {
-    parsedUrl.port = parsedUrl.protocol === 'https' ? '443' : '80';
+    return buildUrlWithStandardPort(parsedUrl);
   }
   return parsedUrl.toString();
 }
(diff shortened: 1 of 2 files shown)


lib/webhooks/webhooks.js
```diff
@@ -3,7 +3,51 @@
...
+  return url;
+}
+
+/**
+ Utility function to add a port number to a URL
+
+ @param {URL} parsedUrl - The parsed url object that Twilio requested on your server
+ @returns {string} - URL with port
+ */
+function addPort(parsedUrl) {
```
thinkingserious: Should we include a parameter to specify a particular user defined port? Or is the function only meant to be used internally?

eshanholtz: Only meant to be used internally.

spec/validation.spec.js
```diff
@@ -75,6 +75,61 @@ describe('Request validation', () => {
...
+
+    it('should validate urls with just username', () => {
+        const urlWithCreds = 'https://user@mycompany.com/myapp.php?foo=1&bar=2';
+        const signature = '2YRLlVAflCqxaNicjMpJcSTgzSs='; // hash of this url
+        const isValid = validateRequest(token, signature, urlWithCreds, defaultParams);
+
+        expect(isValid).toBeTruthy();
+    });
+
+    it('should validate urls with credentials by adding port', () => {
```
thinkingserious: This appears to be the same as the `should validate urls with credentials` test.

eshanholtz: This one ensures standard ports are added to the URL for validation. Note the expected signatures of these two tests are different, but they both pass.

lib/webhooks/webhooks.js
```diff
@@ -3,7 +3,51 @@
...
+
+/**
+ * Utility function to construct the URL string, since Node.js url library won't include standard port numbers
+ *
+ * @param {URL} parsedUrl - The parsed url object that Twilio requested on your server
+ * @returns {string} - URL with standard port number included
+ */
+function buildUrlWithStandardPort(parsedUrl) {
+  let url = '';
+  const port = parsedUrl.protocol === 'https:' ? ':443' : ':80';
```
childish-sambino: Should add a case for the gopher protocol (:
https://nodejs.org/api/url.html#url_url_port

PR #496, opened 2019-11-06
Allow requests to be proxied via TWILIO_PROXY_SERVER

cooperka:
### Motivation

We need to proxy all external requests through a central server in order to comply with certain regulations.

### Changes

If defined, use the `TWILIO_PROXY_SERVER` environment variable for Twilio API requests, otherwise no change.

Happy to tweak this further or change the name of the variable, just let me know. Cheers.

### **Contributing to Twilio**

### DIFF
--- a/lib/rest/Api.js
+++ b/lib/rest/Api.js
@@ -70,7 +70,7 @@ var V2010 = require('./api/V2010');  /* jshint ignore:line */
  */
 /* jshint ignore:end */
 function Api(twilio) {
-  Domain.prototype.constructor.call(this, twilio, 'https://api.twilio.com');
+  Domain.prototype.constructor.call(this, twilio, process.env.TWILIO_PROXY_SERVER || 'https://api.twilio.com');
 
   // Versions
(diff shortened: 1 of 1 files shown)


childish-sambino:
Feels like the better spot for this is in here: https://github.com/twilio/twilio-node/blob/master/lib/base/RequestClient.js

That way it gets applied to _all_ requests if using this client.

Also, I'm not sure how the proxy would know which subdomain to proxy the request to. For instance, in the current change it only applies to `api.twilio.com`, but there are other subdomains such as `serverless.twilio.com`.

cooperka:
Thanks for the feedback @childish-sambino, you're right this only works for `api.twilio.com` and nothing else. That's all we personally needed, so I just grepped for that one; I understand a PR here would need to be more generic.

I think a different change to modify `opts.uri` in `RequestClient.js` as hinted would be more appropriate, but we don't have time to make that change ourselves -- I'll close this since it's unlikely to have progress soon. Cheers.

PR #508, opened 2019-11-21
fix: Rename child TwiML methods to be based on tag name and deprecate old methods

eshanholtz:


lib/twiml/VoiceResponse.d.ts
```diff
@@ -723,6 +776,17 @@ declare namespace VoiceResponse {
...
+  /**
+   * Options to pass to sayAs
+   *
+   * @property interpret-as - Specify the type of words are spoken
+   * @property role - Specify the format of the date when interpret-as is set to date
+   */
+  export interface SsmlSayAsAttributes {
+    "interpret-as"?: SsmlSayAsInterpretAs;
+    role?: SsmlSayAsRole;
+  }
```
childish-sambino: This is duplicated above.

eshanholtz: Fixed.

PR #513, opened 2019-11-27
Add getExpectedTwilioSignature to the index exports

BrandonLegault:
Export the `getExpectedTwilioSignature` function to the main twilio object.

We've added this because in our org we've deployed a service to a FaaS platform so express is not able to pass our application the correct URL in its request object. This will allow us to match the URL manually that Twilio is expecting in its signature.

**Contributing to Twilio**


childish-sambino:
`validateExpressRequest` accepts a `url` as part of its options. Would that suffice?

BrandonLegault:
Apologies, I didn't see that parameter. That works as intended.

PR #540, opened 2020-03-02
[DO NOT MERGE] chore: migrate http client from request to got

eshanholtz:


package.json
```diff
@@ -25,11 +25,12 @@
...
     "lodash": "^4.17.15",
     "q": "2.0.x",
-    "request": "^2.88.2",
+    "got": "^10.6.0",
     "rootpath": "^0.1.2",
     "scmp": "^2.1.0",
     "url-parse": "^1.4.7",
-    "xmlbuilder": "^13.0.2"
+    "xmlbuilder": "^13.0.2",
+    "query-string": "^6.11.1"
```
childish-sambino: Sort plez

lib/base/RequestClient.js
```diff
@@ -45,13 +47,17 @@ RequestClient.prototype.request = function(opts) {
     headers.Authorization = 'Basic ' + b64Auth;
   }
 
+  if (opts.forever) {
+    var keepAliveAgent = new Agent({keepAlive: true});
```
childish-sambino: I don't like this scoping. Prefer something like: `agent: opts.forever === true ? new Agent({keepAlive: true}) : undefined `

package.json
```diff
@@ -25,7 +25,8 @@
     "jsonwebtoken": "^8.5.1",
     "lodash": "^4.17.15",
     "q": "2.0.x",
-    "request": "^2.88.2",
+    "query-string": "^6.11.1",
+    "got": "^10.6.0",
```
childish-sambino: Sort plez

eshanholtz:
Choosing new http client to migrate to since `got` requires deprecating Node 8.

PR #554, opened 2020-04-01
fix: json parsing for non default clients

eshanholtz:


lib/base/Page.js
```diff
@@ -156,7 +156,11 @@ Page.prototype.processResponse = function(response) {
     throw new RestException(response);
   }
 
-  return response.body;
+  try {
+    return JSON.parse(response.body);
+  } catch (e) {
```
thinkingserious, line 161: How about catching the specific exception type here?

PR #601, opened 2020-08-05
Fix: Typescript definition of PhoneNumberCapabilities

ianserlin:
# Fixes #

Fixes a mismatch between the TypeScript definitions, which were lower-cased, and the actual response from the `availablePhoneNumber` api, which has upper-cased `MMS` and `SMS` properties in the `capabilities` field.

An alternative fix is to lowercase these responses from the API before passing them to client code.

Encountered while calling `.availablePhoneNumbers(...).local.list(...)`

Expected response from API (based on Typescript definition):
```js
{
    friendlyName: '(213) 464-xxxx',
    phoneNumber: '+1213464xxx',
    lata: null,
    locality: null,
    rateCenter: null,
    latitude: null,
    longitude: null,
    region: 'CA',
    postalCode: null,
    isoCountry: 'US',
    addressRequirements: 'none',
    beta: false,
    capabilities: { voice: true, sms: true, mms: true, fax: true }
  }
```

Actual response from API (now reflected in update Typescript definition){

```js
{
    friendlyName: '(213) 464-xxxx',
    phoneNumber: '+1213464xxxx',
    lata: null,
    locality: null,
    rateCenter: null,
    latitude: null,
    longitude: null,
    region: 'CA',
    postalCode: null,
    isoCountry: 'US',
    addressRequirements: 'none',
    beta: false,
    capabilities: { voice: true, SMS: true, MMS: true, fax: true }
  }
```

* All unit tests continue to pass without change.
* No additional documentation or inline comments needed.

### Checklist
- [x] I have read the [Contribution Guidelines](CONTRIBUTING.md) and my PR follows them
- [x] I have added necessary documentation about the functionality in the appropriate .md file

### DIFF
--- a/lib/interfaces.d.ts
+++ b/lib/interfaces.d.ts
@@ -5,8 +5,8 @@ export type Url = string;
 export type PhoneNumber = string;
 
 export type PhoneNumberCapabilities = {
-  mms: boolean;
-  sms: boolean;
+  MMS: boolean;
+  SMS: boolean;
   voice: boolean;
   fax: boolean;
 }


ianserlin:
Understood, so one or the other endpoint definition is currently incorrect.

Is it an acceptable strategy to lowercase the incoming data for all requests with `PhoneNumberCapabilities` or is that beyond the scope of what the codegen tool can do?

childish-sambino:
That sounds reasonable to me. We do type-specific deserialization for other types (e.g., [here](https://github.com/twilio/twilio-node/blob/cd53f61f01749616518ebb6a7533625e7c5841c5/lib/rest/api/v2010/account/availablePhoneNumber/local.js#L548)).

So I think just adding a new deserializer for the `PhoneNumberCapabilities` type and reference it on lines like [this](https://github.com/twilio/twilio-node/blob/cd53f61f01749616518ebb6a7533625e7c5841c5/lib/rest/api/v2010/account/availablePhoneNumber/local.js#L555) would work.

childish-sambino:
Closing due to inactivity. Please ping here if work is to continue.

PR #618, opened 2020-10-13
feat: add http logging for Node JS

JenniferMah:
https://issues.corp.twilio.com/browse/DI-947

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,8 +46,8 @@ If your environment requires SSL decryption, you can set the path to CA bundle i
 var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.twilio.com/console
 var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
 
-const client = require('twilio')(accountSid, authToken, { 
-    lazyLoading: true 
+const client = require('twilio')(accountSid, authToken, {
+    lazyLoading: true
 });
 ```
 
@@ -59,7 +59,7 @@ To take advantage of Twilio's [Global Infrastructure](https://www.twilio.com/doc
 var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.twilio.com/console
 var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
 
-const client = require('twilio')(accountSid, authToken, { 
+const client = require('twilio')(accountSid, authToken, {
     region: 'au1',
     edge: 'sydney',
 });
@@ -74,6 +74,22 @@ client.edge = 'sydney';
 ```
 
 This will result in the `hostname` transforming from `api.twilio.com` to `api.sydney.au1.twilio.com`.
+
+### Enable Debug Logging
+There are two ways to enable debug logging in the default HTTP client. You can create an environment variable called `TWILIO_DEBUG` and set it to `true` or you can set the debug variable on the client as true:
+```javascript
+var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.twilio.com/console
+var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
+
+const client = require('twilio')(accountSid, authToken, {
+  debug: true
+});
+```
+You can also set the debug variable on the client as true after constructing the Twilio client:
+```javascript
+const client = require('twilio')(accountSid, authToken);
+client.debug = true;
+```
 
 ## Handling Exceptions
 
--- a/README.md
+++ b/README.md
@@ -76,19 +76,19 @@ client.edge = 'sydney';
 This will result in the `hostname` transforming from `api.twilio.com` to `api.sydney.au1.twilio.com`.
 
 ### Enable Debug Logging
-There are two ways to enable debug logging in the default HTTP client. You can create an environment variable called `TWILIO_DEBUG` and set it to `true` or you can set the debug variable on the client as true:
+There are two ways to enable debug logging in the default HTTP client. You can create an environment variable called `TWILIO_LOG_LEVEL` and set it to `debug` or you can set the logLevel variable on the client as debug:
 ```javascript
 var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.twilio.com/console
 var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
 
 const client = require('twilio')(accountSid, authToken, {
-  debug: true
+  logLevel: "debug"
 });
 ```
-You can also set the debug variable on the client as true after constructing the Twilio client:
+You can also set the logLevel variable on the client after constructing the Twilio client:
 ```javascript
 const client = require('twilio')(accountSid, authToken);
-client.debug = true;
+client.logLevel = "debug";
 ```
 
 ## Handling Exceptions
--- a/README.md
+++ b/README.md
@@ -82,13 +82,13 @@ var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.tw
 var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
 
 const client = require('twilio')(accountSid, authToken, {
-  logLevel: "debug"
+  logLevel: 'debug'
 });
 ```
 You can also set the logLevel variable on the client after constructing the Twilio client:
 ```javascript
 const client = require('twilio')(accountSid, authToken);
-client.logLevel = "debug";
+client.logLevel = 'debug';
 ```
 
 ## Handling Exceptions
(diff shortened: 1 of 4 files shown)


lib/rest/Twilio.js
```diff
@@ -125,6 +126,7 @@ function Twilio(username, password, opts) {
   }
   this.edge = opts.edge || env.TWILIO_EDGE;
   this.region = opts.region || env.TWILIO_REGION;
+  this.debug = opts.debug || (/true/i).test(env.TWILIO_DEBUG);
```
eshanholtz, line 129: Not sure what the second half is doing. Should just be able to have `this.debug = opts.debug || env.TWILIO_DEBUG`

lib/base/RequestClient.js
```diff
@@ -94,6 +95,11 @@ RequestClient.prototype.request = function(opts) {
   this.lastRequest = new Request(optionsRequest);
 
   axios(options).then((response) => {
+    if (opts.debug === true) {
```
eshanholtz, line 98: can simplify to `if (opts.debug)`

lib/base/RequestClient.js
```diff
@@ -94,6 +95,11 @@ RequestClient.prototype.request = function(opts) {
...
+      const requestLog = `Request: ${response.config.method.toUpperCase()} | ${response.config.url} | ${JSON.stringify(response.config.data)} | ${JSON.stringify(response.config.headers)}`
```
eshanholtz, line 99: maybe try writing a helper function and match the style to the CLI? https://github.com/twilio/twilio-cli-core/blob/8748ddb3d1158ffcff40e3fe022977574f301cb4/src/services/cli-http-client.js#L118

lib/rest/Twilio.js
JenniferMah: `env.TWILIO_DEBUG` is originally a string. `(/true/i).test `is using regex to see if `env.TWILIO_DEBUG` includes the value true. It will return a boolean of true or false depending on if `env.TWILIO_DEBUG` includes true. If it doesn't get converted to a boolean here then anytime the user has `env.TWILIO_DEBUG` defined debugger logs will be outputted.

lib/base/RequestClient.js
```diff
@@ -107,4 +111,33 @@ RequestClient.prototype.request = function(opts) {
...
+    console.log(options.data);
+  }
+
+  if (options.params && Object.keys(options.params).length > 0) {
+    console.log('Querystring:');
+    console.log(options.params);
+  }
+
+  const customHeaders = Object.keys(options.headers).filter((header) => {
+    return !STANDARD_HEADERS.includes(header.toLowerCase());
```
childish-sambino, line 131: I don't think we need to remove standard headers. They could be helpful for debugging.

lib/base/RequestClient.js
```diff
@@ -94,6 +95,9 @@ RequestClient.prototype.request = function(opts) {
   this.lastRequest = new Request(optionsRequest);
 
   axios(options).then((response) => {
+    if (opts.debug) {
+      debugLogRequest(response)
```
childish-sambino, line 99: Should also look at logging the response.

README.md
```diff
@@ -75,6 +75,22 @@ client.edge = 'sydney';
 
 This will result in the `hostname` transforming from `api.twilio.com` to `api.sydney.au1.twilio.com`.
 
+### Enable Debug Logging
+There are two ways to enable debug logging in the default HTTP client. You can create an environment variable called `TWILIO_DEBUG` and set it to `true` or you can set the debug variable on the client as true:
```
childish-sambino, line 79: What about allowing different log levels. Like `TWILIO_LOG_LEVEL=(debug/info/warn/error/none)`

lib/base/RequestClient.js
JenniferMah: In addition to the parsed response in `debugLogRequest` also logging the whole response for users like: `console.log(response)`?

childish-sambino: Wait, I was confused. This is called `debugLogRequest` but it's logging the response. That should be fixed.

I meant why not log the request _and_ the response?

CHANGES.md
```diff
@@ -1,6 +1,28 @@
 twilio-node changelog
 =====================
 
+[2020-10-14] Version 3.50.0
```
childish-sambino, line 4: Need to rebase in order to fix these merge issues.

README.md
```diff
@@ -75,6 +75,22 @@ client.edge = 'sydney';
...
 This will result in the `hostname` transforming from `api.twilio.com` to `api.sydney.au1.twilio.com`.
 
+### Enable Debug Logging
+There are two ways to enable debug logging in the default HTTP client. You can create an environment variable called `TWILIO_LOG_LEVEL` and set it to `debug` or you can set the logLevel variable on the client as debug:
+```javascript
+var accountSid = process.env.TWILIO_ACCOUNT_SID; // Your Account SID from www.twilio.com/console
+var authToken = process.env.TWILIO_AUTH_TOKEN;   // Your Auth Token from www.twilio.com/console
+
+const client = require('twilio')(accountSid, authToken, {
+  logLevel: "debug"
```
childish-sambino, line 85: nit: use single-quotes

lib/base/RequestClient.js
```diff
@@ -94,6 +95,15 @@ RequestClient.prototype.request = function(opts) {
   this.lastRequest = new Request(optionsRequest);
 
   axios(options).then((response) => {
+    if (opts.logLevel === "debug") {
+      parseRequest(optionsRequest)
```
childish-sambino, line 99: Should log the actual request (just `options`) and do this _before_ making the request in case of weird errors n such.

lib/base/RequestClient.js
```diff
@@ -94,6 +95,15 @@ RequestClient.prototype.request = function(opts) {
...
+      console.log(`response.statusCode: ${response.status}`)
+      console.log(`response.headers: ${JSON.stringify(response.headers)}`)
+      if (response.status < 200 || response.status >= 400) {
```
childish-sambino, line 102: I think we do this test elsewhere and should not be testing here.

lib/base/RequestClient.js
```diff
@@ -107,4 +117,27 @@ RequestClient.prototype.request = function(opts) {
...
+    console.log('Querystring:');
+    console.log(options.params);
+  }
+
+  if (options.headers) {
+    console.log('Headers:');
+    console.log(options.headers)
+  }
+
+  console.log(`User-Agent: ${options.headers['User-Agent']}`);
```
childish-sambino, line 139: This would be included in the header logging above.

lib/base/RequestClient.js
```diff
@@ -107,4 +117,27 @@ RequestClient.prototype.request = function(opts) {
...
+function parseRequest(options) {
+  console.log('-- BEGIN Twilio API Request --');
+  console.log(`${options.method} ${options.url}`);
+
+  if (options.data) {
+    console.log('Form data:');
+    console.log(options.data);
+  }
+
+  if (options.params && Object.keys(options.params).length > 0) {
```
childish-sambino, line 129: What's the advantage of checking the key length here but not elsewhere?

JenniferMah: Ops! This was originally left in when this function was parsing the response but it is no longer needed.

lib/base/RequestClient.js
JenniferMah: Is there another location that should be outputting this error message or should this just do a quick check to see if `response.data.code || response.data.message || response.data.more_info` exists then print the error message?

childish-sambino: What about just logging `response.data`?

lib/rest/Twilio.js
```diff
@@ -220,6 +222,7 @@ function Twilio(username, password, opts) {
  * @param {object} [opts.data] - The request data
  * @param {int} [opts.timeout] - The request timeout in milliseconds
  * @param {boolean} [opts.allowRedirects] - Should the client follow redirects
+ * @param {string} [opts.logLevel] - Show debug logs
```
childish-sambino, line 225: This isn't a request-level parameter, but maybe it should be: `logLevel: opts.logLevel || this.logLevel`?

JenniferMah: Is this not what's happening [below](https://github.com/twilio/twilio-node/pull/618/files#diff-3685a5bef74c26aa2f3cfd3dae3f1a63557c169fab314b0d342698195782a320R271) ? Is the ask here just to change the parameter comment to state `@param {string} [this.logLevel] - Show debug logs`

childish-sambino: No, what you're doing below is `logLevel: this.logLevel`. You're not using `opts.logLevel` anywhere.

lib/rest/Twilio.js
```diff
@@ -268,7 +268,7 @@ Twilio.prototype.request = function request(opts) {
     data: opts.data,
     timeout: opts.timeout,
     allowRedirects: opts.allowRedirects,
-    logLevel: this.logLevel
+    logLevel: this.logLevel || opts.logLevel
```
childish-sambino, line 271: `opts.logLevel` should take precedence over `this.logLevel`

childish-sambino (approved):

PR #621, opened 2020-10-21
fix: X-Twilio-Signature validation when URL query parameters contain @ or :

adamj9431:

X-Twilio-Signature validation appears to be broken in cases when Twilio sends query parameters containing `@` or `:`. This affects SIP calls since GET calls (e.g. audio URLs) end up containing query parameters that look like `...&Caller=sip%3A1234567890%40209.182.172.181&...` 

It seems the signature header is calculated on the Twilio side without percent-escaping `@` or `:`, but the actual request is then made by Twilio's infrastructure with those characters escaped.

This PR fixes the issue for me, but the other (potentially more robust) fix is to ensure that Twilio infrastructure calculates the signature on the actual query string that is sent to the server after any escaping has occurred.

Note that, [per RFC 3986](https://www.456bereastreet.com/archive/201008/what_characters_are_allowed_unencoded_in_query_strings/) `@` and `:` do not need to be escaped in the query portion of the URI. They do need to be escaped in other parts of the URI, though. Go figure.


thinkingserious:
Hello @adamj9431,

Thank you for taking the time to share your fix! Could you please check the box next to "I acknowledge that all my contributions will be made under the project's license" that I've just added to the description?

With best regards,

Elmer

adamj9431:
@thinkingserious done

childish-sambino (approved):

PR #622, opened 2020-10-21
fix: remove request auth headers from debug logging

JenniferMah:
# Fixes #
Remove request auth header from debugging logs.

### Checklist


lib/base/RequestClient.js
```diff
@@ -133,7 +133,10 @@ function logRequest(options) {
 
   if (options.headers) {
     console.log('Headers:');
-    console.log(options.headers)
+    const headers = Object.keys(options.headers).filter((header) => {
```
thinkingserious, line 136: I think we might want to add a test here.

lib/base/RequestClient.js
```diff
@@ -133,7 +133,10 @@ function logRequest(options) {
 
   if (options.headers) {
     console.log('Headers:');
-    console.log(options.headers)
```
childish-sambino, line 136: Also don't log `options.data` or `response.data`.

lib/base/RequestClient.js
JenniferMah: Should it be a unit test in the Request.Client.spec? and should the test be comparing the logging output or just checking that `headers` does not contain any auth headers?

thinkingserious: I like a unit test in `Request.Client.spec`, minimally checking that `headers` does not contain any forbidden data. Thanks!

spec/unit/base/RequestClient.spec.js
```diff
@@ -46,6 +46,11 @@ describe('lastResponse and lastRequest defined', function() {
     expect(client.lastResponse.body).toEqual('voltron');
   });
 
+  it('should not include request authorization header in filtered headers', function () {
+    const filteredHeaderKeys = client.filterLoggingHeaders(client.lastRequest.headers);
+    expect(filteredHeaderKeys).toEqual(['test-header-key', 'Connection'])
```
childish-sambino, line 51: There's no authorization header to begin with. Should add this to the test data.

JenniferMah: `client.lastRequest.headers` includes the test-header-key, connection, and authorization header when it mocks the request. Is there a additional authorization header that needs to be tested that does not take the format of `Authorization: 'Basic abcd'
` that I need to include in the test data?

childish-sambino: Ah, never mind then.

childish-sambino (approved):

PR #653, opened 2021-02-03
fix: adds property types for AccessToken and Grants

philnash:
The docs show things like adding an identity to an access token like this:

```js
const accessToken = new AccessToken(
  ACCOUNT_SID,
  API_KEY_SID,
  API_KEY_SECRET,
);

accessToken.identity = 'example';
```

And similar for options in grants. But the types don't support this and enforce TS users to set them in the options passed to the constructor.

This adds sthe types for the properties for AccessTokens and each *Grant.

### Checklist

### DIFF
--- a/lib/jwt/AccessToken.d.ts
+++ b/lib/jwt/AccessToken.d.ts
@@ -1,6 +1,11 @@
-declare class AccessToken {
+declare class AccessToken implements AccessToken.AccessTokenOptions {
   static DEFAULT_ALGORITHM: 'HS256';
   static ALGORITHMS: ['HS256', 'HS384', 'HS512'];
+
+  ttl: number;
+  identity?: string;
+  nbf?: number;
+  region?: string;
 
   /**
    * Creates new AccessToken instance
@@ -53,7 +58,11 @@ declare namespace AccessToken {
     TaskRouterGrantOptions,
     TaskRouterGrantPayload,
     'task_router'
-  > {}
+  > implements TaskRouterGrantOptions {
+    workspaceSid?: string;
+    workerSid?: string;
+    role?: string;
+  }
 
   export interface ChatGrantOptions {
     serviceSid?: string;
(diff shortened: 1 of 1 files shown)


thinkingserious (commented):
Thanks for the PR @philnash, just a couple of questions:

1. We noticed that you are using the proper camel casing, but that convention is not consistent with the snake casing we have. Would you recommend updating the snake cased variables to camel casing for consistency and correctness? If so, would that break folks?
2. Thoughts on testing these changes?

philnash:
> We noticed that you are using the proper camel casing, but that convention is not consistent with the snake casing we have. Would you recommend updating the snake cased variables to camel casing for consistency and correctness? If so, would that break folks?

I haven't actually changed anything with regards to the camel casing of the options you can add to access tokens or grants. All I'm doing is exposing the properties' types. Before, to set the properties in TypeScript you needed to do so in the constructor:

```typescript
const chatGrant = new ChatGrant({ serviceSid: "CHxxxx" });
// trying to set a property would cause a type error
chatGrant.pushCredentialSid = "abc123"
// error Property 'pushCredentialSid' does not exist on type 'ChatGrant'
```

However, since these are properties, this works fine in JavaScript (and is how a number of things are documented).

Adding the properties to each grant and the access token mean that you can now set them as properties without TypeScript complaining.

The snake casing in the actual token/grants is actually handled when you create the JWT. For example, this is [the chat grant changing camel case properties to snake case for the token](https://github.com/twilio/twilio-node/blob/main/lib/jwt/AccessToken.js#L55-L66). Like I said, I haven't changed anything here, just exposed these properties as types on the token/grants.

> Thoughts on testing these changes?

It seems that the only testing of TypeScript types is to ensure that [this example compiles](https://github.com/twilio/twilio-node/blob/main/examples/typescript/example.ts). Do you think that this should be added there? Or is there more of a plan for testing the TypeScript types of this repo?

thinkingserious:
Thanks for the detailed response @philnash, makes sense! I think we can leave the testing out of scope for this particular PR.

thinkingserious (approved):
👍🏼

PR #665, opened 2021-03-12
fix: fix validateExpressRequest issue

### LINKED ISSUE #657: validateExpressRequest not working correctly
### Issue Summary
There is a duality in the validator code for [validateExpressRequest](https://github.com/twilio/twilio-node/blob/ffbace7dd3605e09d5861c9e2488e6bdb8c6115b/lib/webhooks/webhooks.js#L147) as [validateRequest](https://github.com/twilio/twilio-node/blob/ffbace7dd3605e09d5861c9e2488e6bdb8c6115b/lib/webhooks/webhooks.js#L178) requires body parameter to be an object, but [validateRequestWithBody](https://github.com/twilio/twilio-node/blob/ffbace7dd3605e09d5861c9e2488e6bdb8c6115b/lib/webhooks/webhooks.js#L171) requires body to be a string. validateRequestWithBody is also called with 
```node
request.body || {}
```
so it fails even when body is not present.

### Steps to Reproduce
1. call validateExpressRequest with a request which has a parsed body object or no body at all

### Exception
If not before, exception occurs in [getExpectedBodyHash](https://github.com/twilio/twilio-node/blob/ffbace7dd3605e09d5861c9e2488e6bdb8c6115b/lib/webhooks/webhooks.js#L84) because body is not a string

### Technical details:
* twilio-node version: 3.56.0
* node version: 15.0.1

shwetha-manvinkurke:
# Fixes #

This PR fixes issue when the `validateExpressRequest` is used for webhook validation when the content-type is application/json. 

Note: We expect the `request.rawBody` populated for the validation in this case. 

Fixes #657 

### Checklist
- [ ] I have added inline documentation to the code I modified

### DIFF
--- a/examples/express.js
+++ b/examples/express.js
@@ -1,0 +1,32 @@
+const twilio = require('twilio');
+const bodyParser = require('body-parser');
+const MessagingResponse = require('twilio').twiml.MessagingResponse;
+
+const express = require('express')
+const app = express()
+const port = 3000
+
+app.use(bodyParser.json({
+    verify: (req, res, buf) => {
+        req.rawBody = buf
+    }
+}))
+
+app.get('/', (req, res) => {
+    res.send('Hello World!')
+})
+
+app.post('/message', twilio.webhook(), (req, res) => {
+    // Twilio Messaging URL - receives incoming messages from Twilio
+    const response = new MessagingResponse();
+
+    response.message(`Your text to me was ${req.body.Body}.
+                    Webhooks are neat :)`);
+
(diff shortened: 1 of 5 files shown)


examples/express.js
```diff
@@ -0,0 +1,32 @@
+const twilio = require('twilio');
```
thinkingserious, line 1: I think it may be worth linking to this example from the README.md.

thinkingserious (approved):
🥇

PR #675, opened 2021-05-26
fix: remove @type/express 

JenniferMah:
# Fixes #
@types/express only supports TypeScript 3.5+ versions while we support TypeScript 2.9+. This PR removes all of the @types dependencies as dev dependencies except for @types/node for environment variables.

### DIFF
--- a/lib/webhooks/webhooks.d.ts
+++ b/lib/webhooks/webhooks.d.ts
@@ -1,3 +1,11 @@
+type Request = (
+  protocol: string,
+  host: string,
+  originalUrl: string,
+  rawBody: Body,
+  body: Body,
+) => any;
+
 export interface RequestValidatorOptions {
   /**
    * The full URL (with query string) you used to configure the webhook with Twilio - overrides host/protocol options
@@ -101,7 +109,7 @@ export function validateRequestWithBody(
  * @param opts - options for request validation
  */
 export function validateExpressRequest(
-  request: any,
+  request: Request,
   authToken: string,
   opts?: RequestValidatorOptions
 ): boolean;
--- a/lib/webhooks/webhooks.d.ts
+++ b/lib/webhooks/webhooks.d.ts
@@ -1,11 +1,3 @@
-import { Request, Response } from 'express';
-
-type Middleware = (
-  request: Request,
-  response: Response,
-  next: () => void
-) => any;
-
 export interface RequestValidatorOptions {
   /**
    * The full URL (with query string) you used to configure the webhook with Twilio - overrides host/protocol options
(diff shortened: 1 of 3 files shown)


lib/webhooks/webhooks.d.ts
```diff
@@ -1,9 +1,9 @@
-import { Request, Response } from 'express';
-
-type Middleware = (
-  request: Request,
-  response: Response,
-  next: () => void
+type Request = (
```
childish-sambino, line 1: Think this should be an `interface` instead which the `express.Request` interface will implicitly implement.

JenniferMah: Something like this? 
```typescript
export interface Request {
  protocol: string;
  host: string;
  originalUrl: string;
  rawBody: Body;
  body: Body;
}
```

childish-sambino: Yes, but no need to export it since client will bring their own types (BYOT?).

`host` should be nested under `headers`, and needs a `header(string)` function.

JenniferMah: Can you clarify what you mean by the host needs a header(string) function? I see that host is nested in headers. Does this mean that I need to get the headers then search for the host?

lib/webhooks/webhooks.d.ts
```diff
@@ -1,10 +1,13 @@
...
-  request: Request,
-  response: Response,
-  next: () => void
-) => any;
+interface Request {
+  protocol: string;
+  host: header;
+  originalUrl: string;
+  rawBody: Body;
+  body: Body;
```
childish-sambino, line 6: Express uses `any` for the `body`. `Body` is ambiguous.

lib/webhooks/webhooks.d.ts
childish-sambino: `header` function needed since we call `request.header('X-Twilio-Signature')`:
```
header(name: string): string | undefined;
```

We also call `request.headers.host` so need a `headers` property:
```
headers: IncomingHttpHeaders;
```

lib/webhooks/webhooks.d.ts
```diff
@@ -1,10 +1,13 @@
...
-  request: Request,
-  response: Response,
-  next: () => void
-) => any;
+interface Request {
+  protocol: string;
+  header(name: string): string | undefined;
+  headers: IncomingHttpHeaders;
+  originalUrl: string;
+  rawBody: any;
```
childish-sambino, line 8: Should be optional since it's artificial: https://github.com/twilio/twilio-node/pull/665
```
rawBody?: any;
```

childish-sambino (approved):
👍

PR #680, opened 2021-07-07
chore: update user agent string for better debugging

philnash:
This PR standardizes the format of our User-Agent string in twilio-node and provides the ability to pass additional information from upstream packages to be included. It follows the agreed pattern of:

`<core-api-lib>/<core-api-lib-version> (<os-name> <os-arch>) node/<node-version> <extensions>`

Where <extensions> are being added space separated following:

`<library>/<library-version>`

or

`<tag>`

In this case the Twilio class allows you to pass in `userAgentExtensions` as an array to specify upstream libraries who use it to add their library name and version.

This could possibly be described as a breaking change as it will update the underlying User-Agent used by the default HTTP Client and anything that relies on that will need an update. I have not marked this as breaking yet though, as I am unsure if that would be a problem and I am interested in other opinions on this.

### Checklist


lib/rest/Twilio.js
```diff
@@ -246,11 +249,21 @@ Twilio.prototype.request = function request(opts) {
...
+
+  var pkgVersion = moduleInfo.version;
+  var osName = os.platform();
+  var osArch = os.arch();
+  var nodeVersion = process.version;
   headers['User-Agent'] = util.format(
-    'twilio-node/%s (node.js %s)',
-    moduleInfo.version,
-    process.version
+    'twilio-node/%s (%s %s) node/%s',
```
thinkingserious, line 258: Should this new standardization be applied across all libraries? (e.g. [twilio-python](https://github.com/twilio/twilio-python/blob/c25b21301c4f692c1e9d76774a906d67be51d175/twilio/rest/__init__.py#L115))

spec/unit/rest/Twilio.spec.js
```diff
@@ -1,87 +1,137 @@
...
+      os.platform(),
+      os.arch(),
+      process.version
+    );
+
+    it('sets the user-agent by default', () => {
+      const client = new twilio('ACXXXXXXXX', 'test-password');
+      const scope = nock('https://api.twilio.com', {
+        reqheaders: {
+          'User-Agent': defaultUserAgent,
```
thinkingserious, line 113: Can we test what this value should be in this case, even if it's just a pattern match?

thinkingserious (approved):
I recommend marking this as a breaking change since the change could break folks.

lib/rest/Twilio.js
childish-sambino: https://twilio.slack.com/archives/CGQPL0RPH/p1625670427257100?thread_ts=1625623548.248000&cid=CGQPL0RPH

spec/unit/rest/Twilio.spec.js
philnash: Fair point, using the exact same code to test the user agent as that which generates it isn't cool.

How about testing it by building a string in a different way to show the expectation a bit clearer?

thinkingserious (approved):
🎆

philnash:
Is there anything else that needs to be done for this PR? I notice that one of the checks that is required needs to be run, how do I do that?

childish-sambino (approved):

PR #720, opened 2022-01-06
feat: add GitHub release step during deploy

childish-sambino:


shwetha-manvinkurke (approved):

.github/workflows/test-and-deploy.yml
```diff
@@ -52,15 +59,14 @@ jobs:
       - name: Get tagged version
         run: echo "GITHUB_TAG=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
 
+      - name: Create GitHub Release
+        uses: sendgrid/dx-automator/actions/release@main
+        env:
```
shwetha-manvinkurke, line 64: Why not add a footer here as well with the npm package link?

childish-sambino: We could, but it's not something we currently do so just trying to match the existing implementation before enhancing.

PR #754, opened 2022-05-24
feat: generate Twilio.ts file

beebzz:
Relates to https://code.hq.twilio.com/twilio/librarian/pull/381


claudiachua (approved):

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,934 @@
+'use strict'
```
claudiachua, line 1: Do we have a header to indicate that this is generated by openapi spec, similar to "this code is generated by yoyodyne"

claudiachua (approved):

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,934 @@
...
+ * @param {string} password -
+ *          The password used for authentication. This is normally auth token, but if using key/secret auth will be
+ *          the secret.
+ * @param {opts} [opts] - The options argument
+ *
+ * @returns {BaseTwilio} A new instance of BaseTwilio
+ */
+/* jshint ignore:end */
+
+class BaseTwilio {
```
childish-sambino, line 54: Should move this into a separate file so we all this hand-crafted code doesn't need to go in the template.

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,934 @@
...
+        this._proxyv1
+        this._ipmessagingv2
+        this._pricingv1
+    }
+
+  }
+}
+
+Object.defineProperty(Twilio.prototype,
+'numbersv2', {
```
childish-sambino, line 411: You should be able to define these from within the class definition itself: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get

lib/rest/Twilio.ts
beebzz: We don't currently do that for OAI-generated files, but would be a good standup/eng sync question

lib/base/BaseTwilio.ts
```diff
@@ -0,0 +1,228 @@
...
+import RequestClient from '../base/RequestClient'; /* jshint ignore:line */
+import { HttpMethod } from '../interfaces'; /* jshint ignore:line */
+
+var os = require('os');  /* jshint ignore:line */
+var url = require('url');  /* jshint ignore:line */
+var moduleInfo = require('../../package.json');  /* jshint ignore:line */
+var util = require('util');  /* jshint ignore:line */
+var RestException = require('../base/RestException');  /* jshint ignore:line */
+
+export interface opts {
```
childish-sambino, line 12: Think this name should be more specific.

lib/base/BaseTwilio.ts
```diff
@@ -0,0 +1,228 @@
...
+    accountSid?: string
+    env?: NodeJS.ProcessEnv
+    edge?: string
+    region?: string
+    lazyLoading?: boolean
+    logLevel?: string
+    userAgentExtensions?: string[]
+}
+
+interface requestOpts {
```
childish-sambino, line 23: Probably should export this.

lib/rest/api/v2010/account/recording.d.ts
```diff
@@ -255,6 +255,7 @@ interface RecordingResource {
   duration: string;
   encryption_details: object;
   error_code: number;
+  media_url: string;
```
childish-sambino, line 258: Shouldn't be pulling in these changes in the PR.

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,518 @@
...
+  _wirelessv1: any
+  _proxyv1: any
+  _ipmessagingv2: any
+  _pricingv1: any
+
+  constructor(username: string, password: string, opts: opts) {
+    super(username, password, opts);
+
+    //Domains
+    this._numbersv2 = undefined
```
childish-sambino, line 124: Don't think this is needed. Should already be `undefined`.

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,518 @@
...
+    this._lookupsv1 = undefined
+    this._autopilotv1 = undefined
+    this._taskrouterv1 = undefined
+    this._wirelessv1 = undefined
+    this._proxyv1 = undefined
+    this._ipmessagingv2 = undefined
+    this._pricingv1 = undefined
+
+    if (!opts.lazyLoading) {
+        this._numbersv2
```
childish-sambino, line 162: These should be calling the non-underscored property.

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,518 @@
...
+        this._chatv1
+        this._conversationsv1
+        this._bulkexportsv1
```
childish-sambino, line 165: Should see about sorting these. Could just need a `sorted()` wrapper on the list return from `listdir`. See the recent Terraform provider changes in librarian for reference.

lib/base/BaseTwilio.ts
```diff
@@ -0,0 +1,228 @@
+'use strict'
+
+import RequestClient from '../base/RequestClient'; /* jshint ignore:line */
+import { HttpMethod } from '../interfaces'; /* jshint ignore:line */
+
+var os = require('os');  /* jshint ignore:line */
```
childish-sambino, line 6: nit: should use `const` over `var` for most cases.

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,481 @@
...
+ *
+ * @returns {Twilio} A new instance of Twilio client
+ */
+/* jshint ignore:end */
+
+
+export class Twilio extends BaseTwilio {
+  username: string
+  password: string
+  opts: clientOpts
```
childish-sambino, line 82: These are defined in the base class.

lib/base/RequestClient.d.ts
```diff
@@ -52,6 +52,10 @@ declare namespace RequestClient {
      * Set to true to use the forever-agent
      */
     forever?: boolean;
+    /**
+     * Set to 'debug' to enable debug logging
+     */
+    logLevel?: string;
```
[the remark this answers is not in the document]
beebzz: When adding the log level to the request as is (without including it in the `RequestClient` datatype), this error pops up: `Object literal may only specify known properties, and 'logLevel' does not exist in type 'RequestOptions<object, object>'`

lib/rest/Twilio.ts
```diff
@@ -0,0 +1,478 @@
...
+  _trusthubv1: any
+  _verifyv2: any
+  _videov1: any
+  _voicev1: any
+  _wirelessv1: any
+
+  constructor(username: string, password: string, opts: ClientOpts) {
+    super(username, password, opts);
+
+    if (!opts.lazyLoading) {
```
childish-sambino, line 120: Should be `this.opts.lazyLoading === false`

childish-sambino (approved):

childish-sambino (approved):

PR #776, opened 2022-08-02
fix: add tsconfig and enable export assignment

childish-sambino:


tsconfig.json
```diff
@@ -0,0 +1,6 @@
+{
+  "compilerOptions": {
+    "target": "es5",
```
beebzz, line 3: Think the target needs to be set to es6 so that the `esModuleInterop` flag works as expected, no?

childish-sambino: It's working on my local with es5.

childish-sambino: It's the typescript tests that are failing. Just trying to get some tests working for the OAI-generator tests to work properly.

beebzz (approved):

PR #779, opened 2022-08-10
test: add test-docker rule

beebzz:


Dockerfile
```diff
@@ -1,12 +1,14 @@
...
 
 ENV NODE_PATH /usr/local/lib/node_modules
 
 COPY lib ./lib
+COPY spec ./spec
 COPY index.js .
 COPY package* ./
 
 RUN npm install . -g
+RUN npm install jasmine -g
```
childish-sambino, line 14: Why does jasmine not get installed as part of the package install?

beebzz: Dev dependencies seem to get ignored when you run a global npm install; think installing normally and adding a `--include=dev` flag makes more sense since it's unclear what we're gaining from doing a global install.

childish-sambino (approved):

PR #791, opened 2022-09-14
chore: drop node versions < 14

beebzz:


.github/workflows/test-and-deploy.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        node: [ 12, 14, 16, lts/* ]
+        node: [ 14, 16, lts/* ]
```
childish-sambino, line 20: Also need to update the min engine version in `package.json`

childish-sambino: Should update the `Dockerfile`, too.

childish-sambino (approved):

PR #794, opened 2022-09-21
feat: add support for node 18

beebzz:


childish-sambino (approved):

.github/workflows/test-and-deploy.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     timeout-minutes: 20
     strategy:
       matrix:
-        node: [ 14, 16, lts/* ]
+        node: [ 14, 16, 18, lts/* ]
```
bnb, line 20: FWIW it might either be worth fully switching to relative syntax (using `*` which is equivalent to current, `lts/*` which is equivalent to the active LTS, and `lts-1`, `lts-2` which are equivalent to one and two majors before the current active LTS, respectively) or only using explicit syntax.

Also, my understanding is that using both `lts/*` and `16` is unnecessarily duplicating runs. I can open a separate PR to change that if you'd like but there's also a suggestion below that would fix it :)

Full relative support:

```suggestion
        node: [ lts/-1, lts/*, * ]
```

Full relative support with further back checks (this will be necessary half of the year, every year):

```suggestion
        node: [ lts/-2, lts/-1, lts/*, * ]
```

Full explicit support:

```suggestion
        node: [ 14, 16, 18 ]
```

bnb: It _does_ look like the `lts/-n` support shipped in [`setup-node@3.3.0`](https://github.com/actions/setup-node/releases/tag/v3.3.0), so we'd want to update that as well on line 28.

bnb: (I'm also happy to leave this all for a totally separate PR, just lmk <3)

childish-sambino: Prefer the full explicit. We look at language version support quarterly to determine what to adopt/drop and don't want to break tests by pulling in the latest version if we haven't tested it yet.

bnb: in that case should `lts/*` be dropped? Currently it's just duplicating compute from 16, and in a few months it'll be duplicating 18.

childish-sambino: @bnb Makes sense. PR welcome!

PR #803, opened 2022-10-12
chore: generate latest with twilio-oai-generator

childish-sambino:


lib/base/Domain.d.ts
```diff
@@ -1,8 +1,9 @@
-import TwilioClient = require('../rest/Twilio');
-import { RequestOpts } from "../base/BaseTwilio";
+import {BaseTwilio, RequestOpts} from "../base/BaseTwilio";
```
claudiachua, line 1: Could be `import {BaseTwilio, RequestOpts } from "./BaseTwilio"; `

childish-sambino: Fixed!

claudiachua (approved):

PR #805, opened 2022-10-20
chore: yeet lodash

beebzz:


lib/base/Domain.js
```diff
@@ -21,7 +20,9 @@ function Domain(twilio, baseUrl) {
  * @return {string} absolute url
  */
 Domain.prototype.absoluteUrl = function(uri) {
-  return _.trim(this.baseUrl, '/') + '/' + _.trim(uri, '/');
+  const cleanUri = uri.replaceAll('/', '');
+  const cleanBaseUrl = this.baseUrl.replaceAll('/','');
```
beebzz, line 24: `replaceAll()` is only supported by Node 15+, which is why Node 14 tests are breaking

lib/base/Domain.js
```diff
@@ -21,7 +20,9 @@ function Domain(twilio, baseUrl) {
...
+  return cleanBaseUrl + '/' + cleanUri;
```
claudiachua, line 25: `/` within the uri/baseUrl will be overwritten. Lodash trim only removes the leading and trailing characters which its not the same as replaceAll.

claudiachua: A potential workaround is to maybe create our own trim function?
```
const trim = (str, c = '\\s') =>
  str.replace(new RegExp(`^([${c}]*)(.*?)([${c}]*)$`), '$2')
```

lib/base/Domain.js
```diff
@@ -31,9 +32,10 @@ Domain.prototype.absoluteUrl = function(uri) {
  * @return {Promise} request promise
  */
 Domain.prototype.request = function(opts) {
-  return this.twilio.request(_.assign({}, opts, {
+  const uri = {
     uri: this.absoluteUrl(opts.uri),
-  }));
+  };
+  return this.twilio.request(Object.assign(opts, Object.getPrototypeOf(opts), uri, Object.getPrototypeOf(uri)));
```
claudiachua, line 38: Is `Object.getPrototypeOf` needed?  Is `Object.assign(opts, uri)` sufficient?

lib/base/RequestClient.js
```diff
@@ -23,159 +22,165 @@ const DEFAULT_TIMEOUT = 30000;
...
+        this.ca = fs.readFileSync(process.env.TWILIO_CA_BUNDLE);
+      }
+      options.ca = this.ca;
+    }
+
+    if (opts.data) {
+      options.data = qs.stringify(opts.data, { arrayFormat: 'repeat' });
+    }
+
+    if (opts) {
```
claudiachua, line 116: Should this be `opts.params`?

lib/base/Version.js
```diff
@@ -31,8 +30,14 @@ Version.prototype.absoluteUrl = function (uri) {
...
  * @return {string} transformed url
  */
-Version.prototype.relativeUrl = function (uri) {
-  return _.trim(this._version, '/') + '/' + _.trim(uri, '/');
+Version.prototype.relativeUrl = function(uri) {
+  if (typeof this._version !== 'string') {
+    if (uri === undefined) {
+      return '';
+    }
+    return uri.trim('/');
```
claudiachua, line 38: I'm assuming this is string.trim function. This function only trims whitespace and does not take in any args. So I don't think this will work...

lib/base/Version.js
```diff
@@ -31,8 +30,14 @@ Version.prototype.absoluteUrl = function (uri) {
...
+  }
+  return this._version.trim('/') + '/' + uri.trim('/');
```
claudiachua, line 40: I'm assuming this is [string.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) function. This function only trims whitespace and does not take in any args. So I don't think this will work.

lib/base/serialize.js
```diff
@@ -98,7 +101,7 @@ module.exports.prefixedCollapsibleMap = function(m, prefix) {
  * @returns {string|object} stringified object
  */
 module.exports.object = function(o) {
-  if (_.isObject(o) || _.isArray(o)) {
+  if (typeof o === 'object' || typeof o === 'array') {
```
claudiachua, line 104: An array is an object so if o is was an array, the `typeof o` would return 'object'.  `if (typeof o === 'object')` should be sufficient

lib/base/serialize.js
```diff
@@ -135,8 +138,8 @@ module.exports.bool = function(input) {
  * @returns {array} new array with transform applied to each element.
  */
 module.exports.map = function(input, transform) {
-  if (_.isArray(input)) {
-    return _.map(input, transform);
+  if (typeof input === 'array' || (typeof input === 'object' && Object.prototype.toString.call(input) === '[object Array]')) {
```
claudiachua, line 141: if `input` was an array, typeof would be `'object'`

lib/http/request.js
```diff
@@ -44,9 +50,8 @@ Request.prototype.toString = function() {
 
   var params = '';
   if (this.params && this.params !== this.ANY) {
-    params = '?' + _.join(_.chain(_.keys(this.params))
-      .map(function(key) { return key + '=' + this.params[key]; }.bind(this))
-      .value(), '&');
+    params = '?' + Object.keys(this.params)
+      .map(key => function() { return key + '=' + this.params[key]; }.bind(this)).join('&');
```
claudiachua, line 54: The function defined in .map is not executed, so each map function returns a function. This will return `?<function>&<function>`. 

fix: add `()` after `.bind(this)` 

`Object.keys(this.params).map(key => function() { return key + '=' + this.params[key]; }.bind(this)()).join('&');`

claudiachua, line 54: The function defined in .map is not executed, so each map function returns a function. This will return ?<function>&<function>.

fix: add () after .bind(this)

Object.keys(this.params).map(key => function() { return key + '=' + this.params[key]; }.bind(this)()).join('&');

claudiachua: `'?' + Object.keys(v).map(key => key + '=' + v[key]).join('&');` this seems like it would work too but I'm unsure if there are corner cases

lib/base/Domain.js
claudiachua, line 38: Is `Object.getPrototypeOf` needed? Maybe just `Object.assign(opts, uri)` is sufficient?

lib/base/Domain.js
claudiachua, line 25: `/` within the uri/baseUrl will be overwritten. Lodash trim only removes the leading and trailing characters which its not the same as `replaceAll`.

A potential workaround is to maybe create our own trim function?

```
const trim = (str, c = '\\s') =>
 str.replace(new RegExp(`^([${c}]*)(.*?)([${c}]*)$`), '$2')
```

```
console.log(trim("/A/BC/DCF/","/"))
=> A/BC/DCF 
```

lib/base/RequestClient.js
claudiachua, line 116: Should this be `opts.params`?

lib/base/Version.js
claudiachua, line 40: I'm assuming this is [string.trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) function. This function only trims whitespace and does not take in any args. So I don't think this will work.

lib/base/serialize.js
claudiachua, line 104: An array is an object so if o is was an array, the `typeof o` would return 'object'. i think `if (typeof o === 'object')` should be sufficient

lib/base/serialize.js
claudiachua, line 141: if `input` was an array, typeof would be 'object' Could remove `typeof input === 'array'`

lib/http/request.js
claudiachua, line 54: The function defined in .map is not executed, so each map function returns a function. This will return `?<function>&<function>`.

fix: add `()` after `.bind(this)`
`
Object.keys(this.params).map(key => function() { return key + '=' + this.params[key]; }.bind(this)()).join('&');
`

claudiachua: `'?' + Object.keys(v).map(key => key + '=' + v[key]).join('&');` this seems like it would work too but I'm unsure if there are corner cases

claudiachua (changes requested):

claudiachua (approved):

lib/base/serialize.js
```diff
@@ -57,19 +56,23 @@ module.exports.iso8601DateTime = function(d) {
  * @return {object} flattened map
  */
 module.exports.prefixedCollapsibleMap = function(m, prefix) {
-  if (_.isUndefined(m) || !_.isPlainObject(m)) {
+  if (!m || typeof m !== 'object' || Object.prototype.toString.call(m) !== '[object Object]') {
```
childish-sambino, line 59: I dislike relying on the string representation. Do we need this for our purposes?

lib/http/request.js
```diff
@@ -55,18 +60,16 @@ Request.prototype.toString = function() {
       data = '\n -G';
     }
 
-    data = data + '\n' + _.join(
-      _.map(this.data, function(value, key) {
-        return ' -d ' + key + '='  + value;
-      }), '\n');
+    data = data + '\n' + Object.entries(this.data).map(d => {
+      return ' -H ' + d[0] + '='  + d[1];
```
childish-sambino, line 64: `-H`?

lib/jwt/AccessToken.js
```diff
@@ -64,48 +58,48 @@ _.extend(ChatGrant.prototype, {
...
 /**
  * @constructor
  * @param {object} options - ...
  * @param {string} options.room - The Room name or Room sid.
  */
-function VideoGrant(options) {
-  options = options || {};
-  this.room = options.room;
-}
+class VideoGrant {
```
childish-sambino, line 69: Is `grant = VideoGrant(opts)` no longer needed?

lib/jwt/ClientCapability.js
```diff
@@ -56,86 +53,84 @@ _.extend(IncomingClientScope.prototype, {
...
-  this.authToken = options.authToken;
-  this.ttl = options.ttl || 3600;
-  this.scopes = [];
-}
+class ClientCapability {
+  constructor(options) {
+    if (!options) {
+      throw new Error('Required parameter "options" missing.');
+    }
+    if (!options.accountSid) {
```
childish-sambino, line 102: Should these be checking against `undefined` instead of negating?

beebzz: `undefined` is falsy by default in JS, so negating it makes it truthy (https://stackoverflow.com/questions/39075142/javascript-undefined-gives-true)

lib/jwt/AccessToken.js
beebzz: Yeah, figured that converting these into proper classes would be easier for us to maintain.

lib/base/serialize.js
beebzz: The string representation is to check that the object being passed in is a "plain object" (since arrays and the like still have the base type of `object`), which is analogous to how lodash already [implements](https://github.com/lodash/lodash/blob/4.17.15/lodash.js#L12036) `isPlainObject()`.

lib/base/Version.js
```diff
@@ -31,8 +30,15 @@ Version.prototype.absoluteUrl = function (uri) {
...
  */
-Version.prototype.relativeUrl = function (uri) {
-  return _.trim(this._version, '/') + '/' + _.trim(uri, '/');
+Version.prototype.relativeUrl = function(uri) {
+  const trim = (str, c = '\\s') => str.replace(new RegExp(`^([${c}]*)(.*?)([${c}]*)$`), '$2');
+  if (typeof this._version !== 'string') {
+    if (uri === undefined) {
+      return '';
+    }
+    return trim(uri, '/');
```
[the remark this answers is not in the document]
beebzz: The original implementation relied on lodash's `trim()` returning empty strings for undefined/null input, will fix it up though

lib/base/deserialize.js
```diff
@@ -85,7 +84,7 @@ function decimal(d) {
  * @return {number} number object
  */
 function integer(i) {
-  return parseNumber(i, _.parseInt);
+  return parseNumber(i, parseInt);
```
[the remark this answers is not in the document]
beebzz: It's native to JS

lib/base/Version.js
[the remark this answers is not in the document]
childish-sambino: Just mean if the first part was null/undefined then the concatenation would have a leading `/`, and if the second part was null/undefined then it would end with a trailing `/`. Not sure if this is important.

childish-sambino (approved):

lib/base/Version.js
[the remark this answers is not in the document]
beebzz: Ah gotcha, fixed

lib/base/Domain.js
```diff
@@ -21,7 +21,20 @@ function Domain(twilio, baseUrl) {
...
+    result += '/';
+  }
+  if (typeof uri === 'string') {
+    uri = trim(uri, '/');
+    if (result === '') {
+      result += '/';
+    }
+    result += uri;
+  }
+  return result;
```
childish-sambino, line 37: Could make this into a `join` function or something that removes the leading/trailing delim from each and then joins them using the delim if both are non-empty strings.

PR #809, opened 2022-10-27
fix: include babel-config and global install twilio for internal testing 

childish-sambino:


Dockerfile
```diff
@@ -8,7 +8,9 @@ ENV NODE_PATH /usr/local/lib/node_modules
 COPY lib ./lib
 COPY spec ./spec
 COPY examples ./examples
-COPY index.js .
-COPY package* ./
+COPY index.js package.json babel.config.js tsconfig.json ./
 
 RUN npm install . --include=dev
+
+RUN npm run build || true # tsc completes but with errors that need to be addressed.
```
claudiachua, line 15: curious on what the usage of "true" in this statement. Can it be `RUN npm run build || #comment...`

claudiachua (approved):

Dockerfile
childish-sambino: `npm run build` currently has a non-zero return value (it fails) so adding a `|| true` enables the image to continue on its merry.

Issue #697, opened 2021-10-07
Taskrouter statistics method expecting SID parameter


serdar:
### Issue Summary
The documentation  to retrieve the statistics for a specific worker (https://www.twilio.com/docs/taskrouter/api/worker/statistics#fetch-a-specific-workers-statistics) states no additional parameters for the `.statistics()` method, however, the current version of the Node library requires an `SID` parameter.

Can the documentation please be updated to refer to the current published version of the Node library, or please specify what the `SID` should be to gather the statistics.

### Code Snippet from Documentation
```node
// Download the helper library from https://www.twilio.com/docs/node/install
// Find your Account SID and Auth Token at twilio.com/console
// and set the environment variables. See http://twil.io/secure
const accountSid = process.env.TWILIO_ACCOUNT_SID;
const authToken = process.env.TWILIO_AUTH_TOKEN;
const client = require('twilio')(accountSid, authToken);

client.taskrouter.workspaces('WSXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
      .workers('WKXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX')
      .statistics()
      .fetch()
      .then(worker_statistics => console.log(worker_statistics.cumulative));
```

### Code Snippet from workerStatistics.d.ts 
```
interface WorkerStatisticsListInstance {
  /**
   * @param sid - sid of instance
   */
  (sid: string): WorkerStatisticsContext;
  /**
   * Constructs a worker_statistics
   */
  get(): WorkerStatisticsContext;
  /**
   * Provide a user-friendly representation
   */
  toJSON(): any;
}
```

### Technical details:
* twilio-node version: 3.67.1
* node version: v16.10.0

eshanholtz:
Ah. This appears to be a bug in our autogenerated code. Thanks for pointing this out, @serdar. This issue has been added to our internal backlog to be prioritized. Pull requests and +1s on the issue summary will help it move up the backlog.

Alex-Bond:
Same problem for `client.messaging.deactivations()`

childish-sambino:
Bumping internally (DI-2172).

Hunga1:
Update: This is now work-in-progress and will be ready for the next release candidate.

PR #820, opened 2022-11-07
fix: use proper parent resource when prefix not matched

childish-sambino:
Fixes https://github.com/twilio/twilio-node/issues/697


rakatyal (approved):

rakatyal:
Changes look fine but SonarCloud failing..

PR #839, opened 2022-12-08
fix: resolve page payload types

beebzz:
[Related PR](https://github.com/twilio/twilio-oai-generator/pull/359)


lib/rest/FlexApiBase.ts
```diff
@@ -11,9 +11,11 @@
 
 import Domain from "../base/Domain";
 import V1 from "./flexApi/V1";
+import V2 from "./flexApi/V2";
 
 class FlexApiBase extends Domain {
   _v1?: V1;
+  _v2?: V2;
```
childish-sambino, line 18: Don't think this should be here.

PR #840, opened 2022-12-09
chore: Rewrite hand-crafted JS files into TS files

Hunga1:
Converts hand-crafted JS files into TS files and removes the accompanying *.d.ts declaration files.

Notable Changes:
- Updated exports/imports in hand-crafted library files and test files to ES module syntax
- Removed TS namespaces (no longer necessary with ES module imports)
- Added CA certificate bundle to `https.Agent` agent options used by `this.axios.defaults.httpsAgent`
- Disabled RequestClient.spec.js `"should cache the CA after loading it for the first time"` test. Current implementation doesn't allow the custom a `https.Agent` to be created at request execution with the current CA bundle for the Axios instance to use (ref: DI-2477)

Internal ref: DI-1418


Hunga1:
Some of the tests will still fail for the following reasons:
- Auto-generated TwiML modules updated to use the TS version of the TwiML module is introduced in pending #838 
- Required `TwilioResponsePayload.uri` property conflicts with optional `uri` property in other interfaces that auto-generated types extend (ref: DI-2475)
- Auto-generated `FlowRevisionInstance._solution` property does not fulfill FlowRevisionContextSolution interface (ref: DI-2475)

Hunga1:
Changes should be merged alongside the removal of the Page namespace references in the generated code (namespace no longer exists). Link to PR: https://github.com/twilio/twilio-node/pull/841

mattcole19 (approved):
LGTM!

spec/unit/base/RequestClient.spec.js
```diff
@@ -284,13 +284,10 @@ describe("lastRequest defined, lastResponse undefined", function () {
 });
 
 describe("User specified CA bundle", function () {
-  let client;
   let options;
   beforeEach(function () {
     axios.create.mockReturnValue(createMockAxios(Promise.reject("failed")));
 
-    client = new RequestClient();
```
claudiachua, line 292: Could we still keep this? `let client = new RequestClient()` instead of initializing it in every test case?

lib/base/Domain.ts
```diff
@@ -0,0 +1,60 @@
...
+    return result;
+  }
+
+  /**
+   * Make request to this domain
+   *
+   * @param {object} opts request options
+   * @return {Promise} request promise
+   */
+  request(opts?: RequestOpts): Promise<any> {
```
claudiachua, line 54: Should `opts` be required? As we are using opts.uri

lib/base/Domain.ts
```diff
@@ -0,0 +1,60 @@
...
+    return this.twilio.request({
+      ...opts,
+      uri: this.absoluteUrl(opts.uri),
```
claudiachua, line 57: Should opts be required? As we are using opts.uri

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+    "last_page_uri",
+    "next_page_uri",
+    "num_pages",
+    "page",
+    "page_size",
+    "previous_page_uri",
+    "start",
+    "total",
+    "uri",
+  ];
```
claudiachua, line 92: Is declaring type META_KEYS[] necessary?

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+  | "first_page_uri"
+  | "last_page_uri"
+  | "next_page_uri"
+  | "num_pages"
+  | "page"
+  | "page_size"
+  | "previous_page_uri"
+  | "start"
+  | "total"
+  | "uri";
```
claudiachua, line 36: Is declaring type `META_KEYS` necessary? Since its being declared again in L80-92

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+    "total",
+    "uri",
+  ];
+
+  /**
+   * Get the url of the previous page of records
+   *
+   * @return {string|undefined} url of the previous page
+   */
+  getPreviousPageUrl(): string {
```
claudiachua, line 99: Should the return type be `string | undefined`?

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    return undefined;
+  }
+
+  /**
+   * Get the url of the next page of records
+   *
+   * @return {string|undefined} url of the next page
+   */
+  getNextPageUrl(): string {
```
claudiachua, line 125: Should the return type be string | undefined?

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+    );
+  }
+
+  /**
+   * Load a list of records
+   *
+   * @param  {object} resources json payload of records
+   * @return {Array} list of resources
+   */
+  loadInstances(resources: TResource[] | TInstance[]): TInstance[] {
```
claudiachua, line 162: Could `resources` be just `TResource[]`? `TInstance[]` was not specify in the previous d.ts

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+    });
+    return instances;
+  }
+
+  /**
+   * Fetch the next page of records
+   *
+   * @return {promise} promise that resolves to next page of results
+   */
+  nextPage(): Promise<Page<TVersion, TPayload, TResource, TInstance>> {
```
claudiachua, line 175: Same as above: Should include `undefined` as return type?

lib/base/Page.ts
```diff
@@ -0,0 +1,278 @@
...
+
+    return nextPagePromise;
+  }
+
+  /**
+   * Fetch the previous page of records
+   *
+   * @return {promise} promise that resolves to previous page of results
+   */
+  previousPage(): Promise<Page<TVersion, TPayload, TResource, TInstance>> {
```
claudiachua, line 201: Same as above: Should include undefined as return type?

lib/base/RequestClient.ts
```diff
@@ -163,13 +254,13 @@ class RequestClient {
       });
   }
 
-  filterLoggingHeaders(headers) {
+  filterLoggingHeaders(headers: Headers | "*") {
```
claudiachua, line 257: What does "*" means?

lib/http/request.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export interface Headers {
+  [header: string]: string;
+}
+
+export default class Request<TData> {
+  method: HttpMethod | "*";
+  url: string;
+  auth: string;
+  params: string;
```
claudiachua, line 23: Should these properties also have `| "*"`

lib/jwt/ClientCapability.ts
```diff
@@ -1,19 +1,38 @@
...
  * @constructor
  * @param filters
  */
-class EventStreamScope {
-  constructor(filters) {
+export class EventStreamScope implements Scope {
+  scope: string = "scope:stream:subscribe";
+  filters: object;
+
+  constructor(filters: object) {
```
claudiachua, line 31: Should `filters` be optional?

lib/jwt/taskrouter/TaskRouterCapability.ts
```diff
@@ -13,8 +44,14 @@ var jwt = require("jsonwebtoken");
...
-class Policy {
-  constructor(options) {
+export class Policy {
+  url?: string;
+  method: string;
+  queryFilter: object;
+  postFilter: object;
+  allow: boolean;
+
+  constructor(options: PolicyOptions) {
```
claudiachua, line 54: Should `options` here be optional?

claudiachua (approved):

lib/base/Page.ts
Hunga1: I don't think that the `META_KEYS` type alias is necessary, but I think it provides the user with more useful information about what the `Page.META_KEYS` static property could be: a list of `META_KEYS` values rather than simply a list of `string` values.

lib/base/Domain.ts
Hunga1: Changed `opts` to required.

lib/base/Page.ts
Hunga1: Done

lib/base/Page.ts
Hunga1: Done

lib/base/Page.ts
Hunga1: Done

lib/base/Page.ts
Hunga1: Done

lib/base/Page.ts
Hunga1: Done

lib/jwt/ClientCapability.ts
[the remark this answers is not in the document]
Hunga1: Changed to optional.

lib/jwt/taskrouter/TaskRouterCapability.ts
Hunga1: Changed to optional.

PR #842, opened 2022-12-15
fix: resolve page payload types

beebzz:
Adds le changes from [here](https://github.com/twilio/twilio-oai-generator/pull/371)


lib/base/Page.ts
```diff
@@ -10,7 +10,7 @@ export interface TwilioResponsePayload {
   page: number;
   page_size: number;
   previous_page_uri: string;
-  uri: string;
+  uri?: string;
```
childish-sambino, line 13: Don't think this is needed.

childish-sambino (approved):

PR #850, opened 2022-12-28
chore: Remove signature overloading

claudiachua:
Internal JIRA Link: [DI-2426] (https://issues.corp.twilio.com/browse/DI-2426)
Related PR: https://github.com/twilio/twilio-oai-generator/pull/376


childish-sambino:
Seeing new tsc errors. E.g.
> Property 'page' does not exist on type 'AwsListInstancePageOptions'

lib/rest/accounts/v1/credential/aws.ts
```diff
@@ -130,10 +130,11 @@ export interface AwsContext {
    * @returns { Promise } Resolves to processed AwsInstance
```
childish-sambino, line 130: Is the `update` overload above here needed?

claudiachua: Yup, I missed that out, it's not needed.

Issue #715, opened 2021-12-16
SMS messaging throws JSON.parse error


homecarecormac:
### Issue Summary
Sending SMS threw error `Unexpected token < in JSON at position 0`.
This doesn't happen always.
This happened only sometimes.
Not sure what makes it happening from time to time.

### Steps to Reproduce
1. Install `twiliow` npm package
2. Configure proper `accountSid` and `authToken`
3. Execute following code snippet

### Code Snippet
```node
const client = require('twilio')(accountSid, authToken);
await client.messages.create({
  body: messageBody,
  to: receipient,
  from: sender,
});
```

### Exception/Log
```
SyntaxError: Unexpected token < in JSON at position 0
```

### Technical details:
* twilio-node version: 3.58.0
* node version: 14

mariomui:
@childish-sambino yeah, i submitted a PR fix for this.
https://github.com/twilio/twilio-node/pull/852/commits/1b4eb8bd05a0edf9d55825d926e6777f6ff4e379
my question would be does the member variables have to be populated on a bad parse? And is there a way to forward the JSON.parse error?

childish-sambino:
Fixed by #855

PR #852, opened 2022-12-29
chore: fix SMS messaging parse error

mariomui:
Chore: fix SMS messaging parse error.

# Fixes #

A short description of what this PR does.

If the JSON.parse ever parses a malformed stringed response , error class will not populate its member variables.
The goal is to always produce an object to populate the member variables or produce a null to flag a no-op.

### Checklist
- [x] I acknowledge that all my contributions will be made under the project's license
- [ ] I have made a material change to the repo (functionality, testing, spelling, grammar)
- [x] I have read the [Contribution Guidelines](https://github.com/twilio/twilio-node/blob/main/CONTRIBUTING.md) and my PR follows them
- [x] I have titled the PR appropriately
- [x] I have updated my branch with the main branch
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added the necessary documentation about the functionality in the appropriate .md file
- [x] I have added inline documentation to the code I modified

If you have questions, please file a [support ticket](https://twilio.com/help/contact), or create a GitHub Issue in this repository.

Issue #570, opened 2020-04-30
Unexpected token U in JSON


meness:
### Issue Summary
It throws an unhandled exception.

### Code Snippet
```node
const twilio = require('twilio')(sid, token);

exports.send = async (to, body) => {
  return twilio.messages.create({
    from: number,
    to,
    body,
  });
};
```

### Exception/Log
```
(node:3376) UnhandledPromiseRejectionWarning: SyntaxError: Unexpected token U in JSON at position 4
    at JSON.parse (<anonymous>)
    at new RestException (C:\Users\Alireza\Desktop\X\node_modules\twilio\lib\base\RestException.js:7:59)
    at success (C:\Users\Alireza\Desktop\X\node_modules\twilio\lib\base\Version.js:135:15)
    at Promise_then_fulfilled (C:\Users\Alireza\Desktop\X\node_modules\q\q.js:766:44)
    at Promise_done_fulfilled (C:\Users\Alireza\Desktop\X\node_modules\q\q.js:835:31)
    at Fulfilled_dispatch [as dispatch] (C:\Users\Alireza\Desktop\X\node_modules\q\q.js:1229:9)
    at Pending_become_eachMessage_task (C:\Users\Alireza\Desktop\X\node_modules\q\q.js:1369:30)
    at RawTask.call (C:\Users\Alireza\Desktop\X\node_modules\asap\asap.js:40:19)
    at flush (C:\Users\Alireza\Desktop\X\node_modules\asap\raw.js:50:29)
    at processTicksAndRejections (internal/process/task_queues.js:75:11)
```

### Technical details:
* twilio-node version: 3.43.0
* node version: 12.10.0

childish-sambino:
I've not been able to recreate this issue. Are you using a custom HTTP client?

meness:
I'm using Axios, but it should not cause a conflict. I tried to wrap Twilio client with a `try/catch` but it still throws the exception. I checked your codes and found you didn't use `async` keyword for the method mentioned in the error, so no one can handle exceptions occur inside of your codes.

childish-sambino:
We recently migrated this library to use `axios` as well so that may be causing an issue, but I'm not able to pin it down. Are you able to share a small example with a custom client that is still breaking?

Regarding catching exceptions, `create` returns a promise so you can either `await` the result and catch exceptions or use `.then(..).catch(...)`.

meness:
[...]

PR #855, opened 2022-12-29
fix: update SMS messaging parse error

mariomui:
Fixes https://github.com/twilio/twilio-node/issues/715

if the messageBody is malformed then there will be a json parse error since the only check to allow parse is whether the body be a string or not. If we wrap the parse in a try catch, we can code a gentler response.

- Assumptions
  - ~~**1: Assuming that details is where the error details are located**~~
    - ~~I have populated the details property inside the exception object with the parse error `{error: bodyParseError}`. For more information, peruse the new unit test described with "should test serialize from improper json string".~~ 
    - on malformed json response_body, the exception should contain undefined properties except for message which is overwritten by the call to super.
  - **2: Assuming that the logger is catching the RestException and not another area of code:**
    - Because this is an intermittent bug, i cannot be fully certain that this PR fixes the parse issue.
    - The lacking string-checking pattern is also found in other locations: 
      - [twilio-node/Version.ts at 5e93c1379fbcb4ea0f7bdc771597755a20950369 · twilio/twilio-node · GitHub](https://github.com/twilio/twilio-node/blob/5e93c1379fbcb4ea0f7bdc771597755a20950369/lib/base/Version.ts#L104)
    - I have contained the scope of this PR to only the RestException

### Checklist

### DIFF
--- a/lib/base/RestException.ts
+++ b/lib/base/RestException.ts
@@ -1,6 +1,13 @@
 "use strict";
+interface RestExceptionError {
+  status: number;
+  message: string;
+  code: number;
+  moreInfo: string;
+  details: object;
+}
 
-export default class RestException extends Error {
+export default class RestException extends Error implements RestExceptionError {
   status: number;
   message: string;
   code: number;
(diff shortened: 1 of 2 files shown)


mariomui:
@childish-sambino as per request within closed PR https://github.com/twilio/twilio-node/pull/852 , i've submitted DI-2442 (with supporting unit test) to target the 4.0.0rc branch instead.

beebzz (approved):
Looks good, thanks for submitting this change!

claudiachua (approved):

lib/base/RestException.ts
```diff
@@ -9,15 +9,39 @@ export default class RestException extends Error {
...
+ */
+function parseResponseBody(response_body) {
+  let body = {};
+  let err = null;
+  try {
+    body = JSON.parse(response_body);
+  } catch (catchError) {
+    const isObject =
+      Object.prototype.toString.call(response_body) === "[object Object]";
+    body = isObject ? response_body : null;
```
childish-sambino, line 43: If the repose body is just plain text like a status message, then it won't make its way into the exception.

lib/base/RestException.ts
```diff
@@ -9,15 +9,39 @@ export default class RestException extends Error {
...
+    if (body) {
+      this.message = body.message;
+      this.code = body.code;
+      this.moreInfo = body.more_info; /* jshint ignore:line */
+      this.details = body.details;
+    }
+    if (body === null) {
+      this.message = "Failed to parse response body";
+      this.code = 12100;
+      this.moreInfo = "https://www.twilio.com/docs/api/errors/12100";
```
childish-sambino, line 24: This does fit right since it's not a 12100 or even necessarily an error produced by Twilio, like in the case of a [451](https://github.com/twilio/twilio-node/issues/570#issuecomment-623019985).

lib/base/RestException.ts
mariomui: okay, i'll just make body = response_body then.

lib/base/RestException.ts
mariomui: i wont need the body null check if the population is confirmed. Will fix.

mariomui: what is the correct code when its a bad json response?
do i still use   "https://www.twilio.com/docs/errors/81022"?

childish-sambino: I don't think either field should be populated if the response is not an object or a JSON string. It likely didn't come from Twilio.

mariomui: @childish-sambino 
based on the commend:
* updated the assumptions written in the PR.
* changed unit test to expect the instanced RestException to contain undefined member variables when the received response_body is a malformed json string 
* simplified the helper function
* introduce a contract to couple the RestException Class to the private helper function. (implements ResponseError)
* Because the exception does not get overwritten when a malformed json response_body string is recieved, the  exception.message retains the supered message: `super("[HTTP " + response.statusCode + "] Failed to execute request")`

childish-sambino (approved):

PR #861, opened 2023-01-03
fix: done only gets called one in each

### LINKED ISSUE #810: done callback is called twice in each() if callback calls done()
### Issue Summary
With at least some flavors of `each`, if `done()` is called manually from the optional `callback`, as suggested by [example.js](https://github.com/twilio/twilio-node/blob/3.83.0/examples/example.js), then the `onComplete` function which calls the `done` optional callback is called twice: once from the optional callback, and once [after finishing](https://github.com/twilio/twilio-node/blob/3.83.0/lib/rest/api/v2010/account.js#L211). I don't see any explicit guarantees in the documentation that `done` is only called once, and it's not a code-breaker if you're aware of it, but I can foresee some undesired outcomes for less-aware developers. It's not immediately obvious to me what the ideal solution is (maybe a different flag from `done` to track whether or not `onComplete` was called?). I've observed the issue in both `calls` (`CallListInstance#each`) and `accounts` (`AccountListInstance#each`), so I'm sure it's quite widespread within a number of yoyodyne files.

Changes may impact the same area as #780 .

### Code Snippet
```node
let n = 0;
twilio.api.v2010.accounts.each({
    callback: (account, done) => {
        console.debug(++n);
        // if account matches test
            done();
    },
    done: console.debug.bind(null, 'done!')
});
```

### Exception/Log
```
1
done! undefined
done! undefined
```

I get similar results for this snippet from the example.js file (for an account with at least 10 calls):
```node
// from examples.js
var i = 0;
// Callback as second parameter
twilio.calls.each({
  pageSize: 7,
  callback: function(call, done) {
    console.log(call.sid);
    i++;
    if (i === 10) {
      done();
    }
  },
  done: function(error) {
    console.log('je suis fini');
    console.log(error);
  }
});
```

### Technical details:
* twilio-node version: ^3.83.0
* node version: v12.13.1

mattcole19:
[Internal Jira Link
](https://issues.corp.twilio.com/browse/DI-2414)
# Fixes #810

Done now only gets called once.

### Checklist

### DIFF
--- a/lib/base/Version.ts
+++ b/lib/base/Version.ts
@@ -260,7 +260,6 @@ export default class Version {
       doneCalled = true;
     }
     function fetchNextPage(fn) {
-      // onComplete();
       let promise = fn();
       if (typeof promise === "undefined") {
         onComplete();
--- a/lib/base/Version.ts
+++ b/lib/base/Version.ts
@@ -242,6 +242,7 @@ export default class Version {
       throw new Error("Callback function must be provided");
     }
     let done = false;
+    let doneCalled = false;
     let currentPage = 1;
     let currentResource = 0;
     let limits = {} as PageLimit;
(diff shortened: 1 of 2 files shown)


childish-sambino (approved):
LGTM, but need to make it pretty.

PR #869, opened 2023-01-04
fix: Handle errors thrown in user callbacks in each

Hunga1:
Fixes issue where errors thrown in user callbacks are not bubbled up. Any errors thrown in user provided `callback` functions in the `each` function will now be caught and passed to the user provided `done` call.

The `each` function will now return a `promise` instead of `void`. The promise will resolve when all records are processed, the limit is reached, or when the user provided `done` function param is executed. The promise will reject if an error occurs in the user callback and the optional user provided `done` function param is not provided, or the user provided `done` function throws an error.

Added unit tests to test functionality of the each function.


Hunga1:
Set to draft until I refactor `each` to always return a promise.

lib/base/Version.ts
```diff
@@ -250,42 +271,69 @@ export default class Version {
...
+              return false;
+            }
+            currentResource++;
+            try {
+              if (callback) {
+                callback(instance, onComplete);
+              }
+            } catch (e) {
+              throw e;
+            }
```
childish-sambino, line 331: Don't think this try-catch is needed since it's just re-throwing.

lib/base/Version.ts
```diff
@@ -250,42 +271,69 @@ export default class Version {
...
+              done ||
+              (typeof params.limit !== "undefined" &&
+                currentResource >= params.limit)
+            ) {
+              done = true;
+              return false;
+            }
+            currentResource++;
+            try {
+              if (callback) {
```
childish-sambino, line 326: Shouldn't a callback always be defined?

childish-sambino (approved):
LGTM, some minor nits.

lib/base/Version.ts
```diff
@@ -229,10 +229,31 @@ export default class Version {
...
+   *  with every request
+   * @param {function} params.callback Callback function to call with each
+   *  record instance
+   * @param {function} [params.done] Optional done function to call when all
+   *  records are processed, the limit is reached, or an error occurs.
+   *  Receives an error argument if an error occurs.
+   * @param {function} [callback] Callback function to call with each record.
+   *  Receives a done function argument that will short-circuit the for-each
+   *  loop that may accept an error argument.
+   * @returns {Promise<void>} Returns a promise that resolves when all records
```
claudiachua, line 249: Raghav made changes to docstring to follow [tsdoc](https://github.com/twilio/twilio-node/blob/4.0.0-rc/lib/jwt/AccessToken.ts#L248) recently. would be nice to adhere to it

spec/unit/base/Version.spec.js
```diff
@@ -151,3 +153,220 @@ describe("done should only be called once in each", () => {
...
+    client = new Twilio("ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX", "AUTHTOKEN", {
+      httpClient: holodeck,
+    });
+  });
+
+  it("should call user callback foreach resource instance", function (done) {
+    let mockCallback = jest.fn();
+    holodeck.mock(new Response(200, bodyOne));
+    holodeck.mock(new Response(200, bodyTwo));
+    holodeck.mock(new Response(200, bodyThree));
```
claudiachua, line 184: Could `holodeck.mock` be inside `beforeEach`, since its repeating in everything testcase

lib/base/Version.ts
Hunga1: Removed

lib/base/Version.ts
Hunga1: Yes, I was narrowing it to avoid a tsc error. It looks like the compiler isn't able to understand that `callback` can never be undefined thanks to the conditional check and exception thrown above. I changed this to use an [optional chaining operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html), which seems to also avoid the compiler error.

lib/base/Version.ts
Hunga1: Done

spec/unit/base/Version.spec.js
Hunga1: Good question. Although most of the unit tests have the same mocked responses, not all of them do and there is no guarantee that future tests in this test group that covers the `each` function may need the same mocked responses. Therefore, I think it's best to limit the setup of each test (in the `beforeEach`) to only variables and states that are needed for all tests.

claudiachua: Sounds good to me, thanks for clarifying!

claudiachua (approved):

PR #872, opened 2023-01-04
chore: Remove signature overloading

claudiachua:
Internal JIRA Link: [DI-2426 ](https://issues.corp.twilio.com/browse/DI-2426)
Related PR: https://github.com/twilio/twilio-oai-generator/pull/376


lib/rest/accounts/v1/credential/aws.ts
```diff
@@ -181,9 +184,14 @@ export class AwsContextImpl implements AwsContext {
...
-  update(params?: any, callback?: any): Promise<AwsInstance> {
+  update(
+    params?:
+      | AwsContextUpdateOptions
+      | ((error: Error | null, item?: AwsInstance) => any),
+    callback?: (error: Error | null, item?: AwsInstance) => any
+  ): Promise<AwsInstance> {
     if (typeof params === "function") {
-      callback = params;
+      callback = params as (error: Error | null, item?: AwsInstance) => any;
```
childish-sambino, line 194: Is this needed? Not seeing any TS warnings w/o it.

lib/rest/accounts/v1/credential/aws.ts
```diff
@@ -400,85 +411,44 @@ export interface AwsListInstance {
...
-   * The request is executed immediately.
-   *
-   * If a function is passed as the first argument, it will be used as the callback
-   * function.
-   *
-   * @param { function } [callback] - Callback to handle list of records
-   */
-  getPage(
-    callback?: (error: Error | null, items: AwsPage) => any
-  ): Promise<AwsPage>;
```
childish-sambino, line 440: Should this have been removed? Can no longer just pass in a callback.

lib/rest/accounts/v1/credential/aws.ts
```diff
@@ -570,11 +529,13 @@ export function AwsListInstance(version: V1): AwsListInstance {
   };
 
   instance.page = function page(
-    params?: any,
-    callback?: any
+    params?:
+      | AwsListInstancePageOptions
+      | ((error: Error | null, item?: AwsPage) => any),
+    callback?: (error: Error | null, item?: AwsPage) => any
   ): Promise<AwsPage> {
```
childish-sambino, line 536: See a TS error here when strict mode is enabled

lib/rest/accounts/v1/credential/aws.ts
```diff
@@ -584,7 +545,7 @@ export function AwsListInstance(version: V1): AwsListInstance {
 
     if (params["pageSize"] !== undefined) data["PageSize"] = params["pageSize"];
 
-    if (params.page !== undefined) data["Page"] = params.pageNumber;
+    if (params.pageNumber !== undefined) data["Page"] = params.pageNumber;
```
childish-sambino, line 548: Good catch.

lib/rest/accounts/v1/credential/aws.ts
claudiachua: It was previously needed for [flowRevision.ts](https://github.com/twilio/twilio-node/actions/runs/3802537470) but the error is now gone. Will removed it

lib/rest/accounts/v1/credential/aws.ts
claudiachua: getPage implementation called [BaseTwilio.ts request](https://github.com/twilio/twilio-node/blob/4.0.0-rc/lib/base/BaseTwilio.ts#L121) and an error will be thrown if user does not pass in `targetURL`. I think that user should not be able to call getPage with just a callback - just to get an error later

lib/rest/accounts/v1/credential/aws.ts
claudiachua: Fixed! Typo with `item?` (should be `items`)

childish-sambino (approved):
🤘

Issue #595, opened 2020-07-08
AccessToken identity option should be required


YassienW:
### Issue Summary
When creating a new AccessToken object, "identity" is an optional parameter of the options object.
https://www.twilio.com/docs/libraries/reference/twilio-node/3.47.0/AccessToken.html

According to the documentation [here](https://www.twilio.com/docs/video/tutorials/user-identity-access-tokens#generate-helper-lib) (and my testing), all access tokens require the following:

- A Twilio Account SID, which is the public identifier of the Twilio account associated with the Access Token.
- An API Key SID, which is the public identifier of the key used to sign the token.
- An Identity grant, which sets the Twilio user identifier for the client holding the token.
- The API Key Secret associated with the API Key SID is used to sign the Access Token and verify that it is associated with your Twilio account.

I believe identity should be required in the constructor, since that is a breaking change the documentation should at least be updated

### Technical details:
* twilio-node version: 3.47.0

eshanholtz:
Nice catch. This issue has been added to our internal backlog to be prioritized. Pull requests and +1s on the issue summary will help it move up the backlog.

rangelfinal:
Lost a lot of time for this issue, was getting a generic "Error 20101: Invalid Access Token" until identity was added to access token options

childish-sambino:
Bumped up the internal backlog (ref: DI-2445).

PR #875, opened 2023-01-06
feat!: Making AccessToken Identity required

rakatyal:
JIRA: [DI-2445](https://issues.corp.twilio.com/browse/DI-2445)
Making Identity required in the AccessToken constructor.

### DIFF
--- a/lib/jwt/AccessToken.ts
+++ b/lib/jwt/AccessToken.ts
@@ -79,9 +79,9 @@ export interface AccessTokenOptions {
    */
   ttl?: number;
   /**
-   * The identity of the first person
+   * The identity of the first person. Required.
    */
-  identity?: string;
+  identity: string;
   /**
    * Time from epoch in seconds for not before value
    */
@@ -329,7 +329,7 @@ export default class AccessToken implements AccessTokenOptions {
   keySid: string;
   secret: string;
   ttl: number;
-  identity?: string;
+  identity: string;
   nbf?: number;
   region?: string;
   grants: Grant<any, any, any>[];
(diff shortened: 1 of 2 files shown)


lib/jwt/AccessToken.ts
```diff
@@ -359,7 +359,9 @@ export default class AccessToken implements AccessTokenOptions {
     if (!secret) {
       throw new Error("secret is required");
     }
-    options = options || {};
+    if (!options || !options.identity) {
```
childish-sambino, line 362: Is this first !-check needed?

spec/unit/jwt/AccessToken.spec.js
```diff
@@ -255,8 +280,9 @@ describe("AccessToken", function () {
     });
 
     it("should create token with multiple grants", function () {
-      var token = new twilio.jwt.AccessToken(accountSid, keySid, "secret");
-      token.identity = "ID@example.com";
+      var token = new twilio.jwt.AccessToken(accountSid, keySid, "secret", {
+        identity: "ID@example.com",
+      });
```
childish-sambino, line 285: Can we just define the token once and use it everywhere, or create it with a function to reduce the duplication?

rakatyal: Not possible to define once since we are overriding properties in tests. Created a function to make it better.

lib/jwt/AccessToken.ts
rakatyal: Actually this breaks if options is undefined. https://github.com/twilio/twilio-node/actions/runs/3857410255/jobs/6574774047

childish-sambino (approved):

rakatyal:
Fixes #595

PR #878, opened 2023-01-09
fix: resolve tsc errs in handcrafted code

beebzz:


childish-sambino (approved):
Could you also remove the `--strict false` flag from the build step?

PR #882, opened 2023-01-11
docs: update upgrade guide with all v4 changes

childish-sambino:


beebzz (approved):
Side note: should we also replace `object` usage w/`Record` in handcrafted code?

PR #886, opened 2023-01-12
feat: Add auto-retry with exponential backoff

Hunga1:
Adds auto-retry with exponential backoff feature to auto-retry requests to the API when the API returns a 429 Too Many Requests error response. This feature is disabled by default.

Adds how-to docs for this feature to the README.

This PR is based on changes made in #741 (Thanks, @bobtfish!)

### DIFF
--- a/lib/base/RequestClient.ts
+++ b/lib/base/RequestClient.ts
@@ -1,5 +1,5 @@
 import { HttpMethod } from "../interfaces";
-import axios, { AxiosInstance, AxiosRequestConfig } from "axios";
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from "axios";
 import * as fs from "fs";
 import HttpsProxyAgent from "https-proxy-agent";
 import qs from "qs";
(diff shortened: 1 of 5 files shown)


lib/base/RequestClient.ts
```diff
@@ -92,13 +95,79 @@ export interface RequestClientOptions {
...
+
+interface ExpontentialBackoffResponseHandlerOptions {
+  /**
+   * Maximum retry delay in milliseconds
+   */
+  maxIntervalMillis: number;
+  /**
+   * Maximum number of request retries
+   */
+  maxRetries: number;
```
childish-sambino, line 129: nit: Should really be named `maxAttempts` since the first attempt is included in the counting.

childish-sambino (approved):
🥇

lib/base/RequestClient.ts
Hunga1: The `maxRetries` will set the maximum number of retries that the client will make when encountering `429` error responses, not including the first attempt.

The response handler function returned by `getExpontentialBackoffResponseHandler` will run on each response received by the RequestClient's axios instance. Although `retryCount` is incremented in [L146](https://github.com/twilio/twilio-node/pull/886/files#diff-c0ee1ee5484bd0fc609af970b02a1d5a19b8bf82fef04d2b09e76c6374383327R146), the function will evaluate if `retryCount <= maxRetries` before setting the `retryCount` value in the `AxiosRequestConfig` and retrying the request in the conditional block. If the `retryCount` exceeds the `maxRetries`, it will skip the conditional block and return the 429 error response.

childish-sambino: Ack, my mistake.

PR #911, opened 2023-02-01
fix: restore test-docker rule

beebzz:
The `test-docker` make rule got lost when we merged in the RC branch, restoring it here


Dockerfile
```diff
@@ -12,5 +12,4 @@ COPY index.* package.json babel.config.js tsconfig.json ./
 
 RUN npm config set unsafe-perm true # Needed to run prepublish as root.
 
-RUN npm install . --include=dev
-RUN npm install -g .
```
childish-sambino, line 16: This line is needed for tricorder snippet testing.

childish-sambino (approved):

PR #916, opened 2023-02-21
test: refactor cluster tests for pack output

childish-sambino:
[DI-2566](https://issues.corp.twilio.com/browse/DI-2566)

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     "express": "^4.17.1",
     "jest": "^29.3.1",
     "jshint": "^2.11.0",
-    "localtunnel": "^2.0.2",
     "mock-fs": "^5.2.0",
     "nock": "^13.2.9",
     "node-mocks-http": "^1.8.1",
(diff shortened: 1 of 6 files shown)


Hunga1 (approved):

package.json
```diff
@@ -64,8 +63,6 @@
     "nsp": "npm audit --production",
     "prettier": "prettier --write .",
     "prettier-check": "prettier --check .",
-    "cluster": "jest ./cluster_test.spec.ts --coverage",
```
rakatyal, line 67: Should we have a command to run cluster test locally without npm packaging?

rakatyal (approved):

rakatyal:
Looks good! Please link the JIRA ticket in the description too.

package.json
childish-sambino: I think it would be nice, but it requires either duplicating the cluster test file or scripting changing the import(s).

PR #1090, opened 2025-05-20
chore: handle free form object using any

### LINKED ISSUE #1076: [Bug] Invalid types in last update
Some of the changes in the following commit define types don't match the actual structure of the data causing type errors when attempting to access values as documented. 

For example, `lineTypeIntelligence: Record<string, object>` suggests that the shape would be:

```
lineTypeIntelligence: {
   key: { ... }
}
```

It should probably instead be `Record<string, any>`. 

Or, even better, it should contain the actual available keys and value types as documented. 

https://github.com/twilio/twilio-node/commit/49f3f2b147818bf03d0067ad7cc13fbcbaf72c7d#diff-55a60b51b684465835a0448f6efceb0ef9b0cb5ada598728e1594ebf2507deddR292

tiwarishubham635:
# Fixes #1076

These are preview changes for handling free form objects correctly using `Record<string, any>` instead of `Record<string, object>`.
Note: Close this PR only when the OAI generator changes are merged.

### Checklist
- [ ] I have made a material change to the repo (functionality, testing, spelling, grammar)
- [ ] I have read the [Contribution Guidelines](https://github.com/twilio/twilio-node/blob/main/CONTRIBUTING.md) and my PR follows them
- [ ] I have titled the PR appropriately
- [ ] I have updated my branch with the main branch
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added inline documentation to the code I modified

### DIFF
--- a/src/rest/bulkexports/v1/export/job.ts
+++ b/src/rest/bulkexports/v1/export/job.ts
@@ -130,7 +130,7 @@ interface JobPayload extends JobResource {}
 interface JobResource {
   resource_type: string;
   friendly_name: string;
-  details: Record<string, object>;
+  details: Record<string, any>;
   start_day: string;
   end_day: string;
   job_sid: string;
@@ -174,7 +174,7 @@ export class JobInstance {
   /**
    * The details of a job which is an object that contains an array of status grouped by `status` state.  Each `status` object has a `status` string, a count which is the number of days in that `status`, and list of days in that `status`. The day strings are in the format yyyy-MM-dd. As an example, a currently running job may have a status object for COMPLETED and a `status` object for SUBMITTED each with its own count and list of days.
    */
-  details: Record<string, object>;
+  details: Record<string, any>;
   /**
    * The start time for the export specified when creating the job
    */
--- a/src/rest/accounts/v1/bulkConsents.ts
+++ b/src/rest/accounts/v1/bulkConsents.ts
@@ -118,7 +118,7 @@ export function BulkConsentsListInstance(
 interface BulkConsentsPayload extends BulkConsentsResource {}
 
 interface BulkConsentsResource {
-  items: Record<string, object>;
+  items: Record<string, any>;
 }
 
 export class BulkConsentsInstance {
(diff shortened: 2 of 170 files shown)


src/rest/bulkexports/v1/export/job.ts
```diff
@@ -174,7 +174,7 @@ export class JobInstance {
   /**
    * The details of a job which is an object that contains an array of status grouped by `status` state.  Each `status` object has a `status` string, a count which is the number of days in that `status`, and list of days in that `status`. The day strings are in the format yyyy-MM-dd. As an example, a currently running job may have a status object for COMPLETED and a `status` object for SUBMITTED each with its own count and list of days.
    */
-  details: Record<string, object>;
+  details: Record<string, any>;
```
msheakoski, line 177: This should be an array, `Record<string, any>[]`. Please see my original comment from when the breaking change was introduced: https://github.com/twilio/twilio-node/issues/1076#issuecomment-2715845955

tiwarishubham635: I have mentioned there that this particular field is a list but in our specification it is written as an object. I'll convey this to my team to fix this. It is still under process

tiwarishubham635: I just checked with my team and I think what they want to send in this field is something of this sort:
```node
{
   [
    {
      status: 'Submitted',
      count: 3,
      days: [ '2025-03-07', '2025-03-08', '2025-03-09' ]
    }
  ]
}
```
Thus it should be object and not list. Is this correct with your API call too?

msheakoski: The API response that I received is an array of objects. The example where you show an array nested directly inside of an object is not valid JavaScript or JSON so that would not be possible to do.

tiwarishubham635: Yeah you are right, let me get this fixed with my team. Thanks!

tiwarishubham635:
no more needed as it is fixed in 5.7.3

PR #1184, opened 2026-03-24
chore: replace scmp with crypto.timingSafeEqual

### LINKED ISSUE #1168: Replace scmp with node crypto.timingSafeEqual()
Warning when deployed in Node 24

npm warn deprecated scmp@2.1.0: Just use Node.js's crypto.timingSafeEqual()

Methuselah96:
# Fixes #1168

scmp is now deprecated and is equivalent to `crypto.timingSafeEqual` (which is supported by Node.js >=6.6.0)

### Checklist

### DIFF
--- a/src/webhooks/webhooks.ts
+++ b/src/webhooks/webhooks.ts
@@ -1,4 +1,3 @@
-const scmp = require("scmp");
 import crypto from "crypto";
 import urllib from "url";
 import { IncomingHttpHeaders } from "http2";
@@ -257,7 +256,10 @@ function validateSignatureWithUrl(
     params
   );
 
-  return scmp(Buffer.from(twilioHeader), Buffer.from(signatureWithoutPort));
+  return timingSafeEqual(
+    Buffer.from(twilioHeader),
+    Buffer.from(signatureWithoutPort)
+  );
 }
 
 export function validateBody(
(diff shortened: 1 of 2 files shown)


Methuselah96:
Investigating failing tests

Methuselah96:
Fixed

src/webhooks/webhooks.ts
```diff
@@ -257,15 +256,23 @@ function validateSignatureWithUrl(
...
   bodyHash: any[] | string | Buffer
 ): boolean {
   var expectedHash = getExpectedBodyHash(body);
-  return scmp(Buffer.from(bodyHash), Buffer.from(expectedHash));
+  return timingSafeEqual(Buffer.from(bodyHash), Buffer.from(expectedHash));
+}
+
+function timingSafeEqual(a: Buffer, b: Buffer) {
+  if (a.length !== b.length) return false;
+  return crypto.timingSafeEqual(a, b);
```
imcotton: Returning earlier would be unsafe because it would leak timing info. Suggest doing the HMAC beforehand:

```js
const salt = crypto.randomBytes(42);
return crypto.timingSafeEqual(
  crypto.createHmac("sha256", salt).update(a).digest(),
  crypto.createHmac("sha256", salt).update(b).digest(),
);
```

Methuselah96: Hmm, I'm out of my depth here, but [`scmp` was returning earlier based on the length](https://github.com/freewil/scmp/blob/master/index.js#L20-L24), so seems like this might be a pre-existing issue

Methuselah96: According to [this comment](https://github.com/nodejs/node/issues/17178#issuecomment-348784606), it's okay to check the length first and return early if they're not equal

Methuselah96: However, according to [this article](https://developers.cloudflare.com/workers/examples/protect-against-timing-attacks/), you should not return early and instead still compare two values. That seems like a more understandable solution to me, so I've updated the PR to that solution.

imcotton: Both negate when lengths differ or HMAC works.

The reason I suggested a later approach is that it is not always clear whether the input is from a or b, thus making it error-prone without context.

_btw: I wrote that code snippet on Cloudflare docs._

Methuselah96: 👍 I've updated the variable names to more clearly indicate which input is the user value and which one is the secret value and verified that we are passing the arguments in the correct order that matches the parameter order

imcotton: Given that it has only been used internally, I think it is good enough to make a cut. LGTM

Thanks.

Emperiusm:
+1 from a downstream consumer.

We hit `npm warn deprecated scmp@2.1.0: Just use Node.js's crypto.timingSafeEqual()` during a transitive-deps audit and confirmed the only path to it is `twilio@5.x → scmp`. v6.0.1 still pulls scmp, so the deprecation can't be cleared by bumping the major. This PR is the only path forward — would love to see it land.

If maintainers want a small validation harness, the existing webhook/auth-token signature validators that route through `scmp` are well-covered by the repo's tests, so a swap-only PR like this one should be low-risk under those.

yinzara:
Now into September with additional releases of this library and still this PR is not merged.  Can someone from @twilio-product-security please look at this?

